PR #10, opened 2020-07-17
[ADD] hr_attendance_user: New module

manuelcalerosolis:
The module allows a user with a role of Manual Attendance to see their own attendance records, and only their own.


hr_attendance_reason/views/hr_attendance_reason_view.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <record id="hr_attendance_reason_view_form" model="ir.ui.view">
+    hr_attendance_reason_view_form" model="ir.ui.view">
```
jarroyomorales, line 3: This diff doesnt belong to the PR and its making the CI red

PR #22, opened 2020-11-27
[14.0][MIG] hr_attendance_modification_tracking

aphon61bank:
I'm edit track_visibility = tracking.


hr_attendance_modification_tracking/tests/test_hr_attendance.py
```diff
@@ -83,7 +83,9 @@ def test_attendance_edit(self):
         dto = datetime.now() + relativedelta(hours=1)
         att.write({"check_out": dto.strftime(DF)})
         self.assertEqual(
-            att.time_changed_manually, True, "Use case 5: Wrong value, data changed."
```
ao-landoo: I don't think changing this assert from True to False is something valid. This shows the behavour has been changed in the migration.

hr_attendance_modification_tracking/tests/test_hr_attendance.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+        # Use case 5:
+        # Change previous attendance check-out to now + 1 hour
+        # Expected: manually_changed = True
+        dto = datetime.now() + relativedelta(hours=1)
+        att.write({"check_out": dto.strftime(DF)})
+        self.assertEqual(
+            att.time_changed_manually,
+            False,
+            "Use case 5: Wrong value, data changed.",
```
ao-landoo, line 88: I don't think changing this assert from True to False is something valid. This shows the behavour has been changed in the migration.

ao-landoo: What do you think @Saran440 ??

Saran440:
Close, Move on https://github.com/OCA/hr-attendance/pull/36

PR #25, opened 2020-12-10
[13.0][FIX] hr_attendance_report_theoretical_time: Fix error in hr.leave form (user without employee group)

victoralmau:
Actually error appear in `hr.leave` form when user without employee group. 
The solution is create `theoretical_hours_start_date` field in `hr.employee.public` model because is the model that users without employee groups can access.

Please, @pedrobaeza and @joao-p-marques review it

@Tecnativa TT27054

### DIFF
--- a/hr_attendance_report_theoretical_time/models/__init__.py
+++ b/hr_attendance_report_theoretical_time/models/__init__.py
@@ -2,6 +2,7 @@
 
 from . import hr_attendance
 from . import hr_employee
+from . import hr_employee_public
 from . import hr_holidays_public
 from . import hr_leave
 from . import hr_leave_type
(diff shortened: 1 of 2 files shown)


joao-p-marques (approved):
LGTM :+1: 

Would it be possible to add a test to cover the issue this is solving?

victoralmau:
> LGTM +1
> 
> Would it be possible to add a test to cover the issue this is solving?

IMO isn't possible to create test because these addon not related with leaves.

joao-p-marques:
Ok, I as just wondering if tests could be improved. It is a good solution
anyway. 👍

pedrobaeza (approved):

pedrobaeza:
Tests are failing

pedrobaeza:
Shouldn't this be on `hr.public.base`?

victoralmau:
> Shouldn't this be on `hr.public.base`?

I had changed before to `hr.employee.base` but don't work with create view (https://github.com/OCA/hr-attendance/blob/13.0/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py#L130-L162) and i hadn't another solution that add field to `hr.employee` and `hr.employee.public` in these case.

Issue #19, opened 2020-11-11
[13.0] hr_attendance_report_theoretical_time: Incompatibility with hr_expense


angelmoya:
You get an error creating expense report with employee without access to private addresses.

Steps to reproduce:
- Create a user without access to private addresses.
- Create employee for this user
- Create a expenses
- In expense form, save and click on "Create report" Button

![2020-11-11 08-01-53](https://user-images.githubusercontent.com/6750341/98780332-e1aaed80-23f4-11eb-975b-31c58640f2cb.png)

Error reason:

In hr standard module we have hr.employee and hr.employee.public, hr.employee.public is a sql view to get readonly acces to employee.

When you create a report with user without private addresses access you spect to get all fields created on hr.employee.

Maybe this is Odoo bug, and not bug on this report, but to prevent this error you can create this fields also in hr.employee.public.

pedrobaeza:
Which OCA module?

angelmoya:
At least:
- hr_attendance_report_theoretical_time, because add theoretical_hours_start_date on hr.employee and not on hr.employee.public
- hr_attendance_rfid, because add rfid_card_code on hr.employee and not on hr.employee.public

will fail with all field (no o2m, m2m or computed no stored) added to hr.employee and not on hr.employee.public

So, there are two sides to this matter:
1. It's a hr odoo module bug
2. It's a hr odoo module non feature, and OCA bug

pedrobaeza:
Then please title properly the issue

pedrobaeza:
They can be OCA bug due to the new model, but the solution is simple: move that fields from `hr.employee` to `hr.employee.public`.

angelmoya:
Ok, thanks Pedro... I know solution si simple... Ok, I will fix, the question is:

Is really needed and well difened that all records present on hr.employee must be present on hr.employee.public?

Or is a Odoo hr_expense module bug?

pedrobaeza:
I haven't dug enough to be able to say, Ángel. Maybe it's not and they have only tested the duality with non relational fields.

pedrobaeza:
Shouldn't it be in `hr.employee.base` instead?

pedrobaeza:
`hr_attendance_report_theoretical_time` fixed in #25

victoralmau:
> `hr_attendance_report_theoretical_time` fixed in #25

[...]

PR #27, opened 2020-12-10
[13.0][IMP] hr_attendance_rfid: Change to hr.employee.base

victoralmau:
Change to `hr.employee.base` model
Related to https://github.com/OCA/hr-attendance/issues/19

@Tecnativa TT27054

### DIFF
--- a/hr_attendance_rfid/security/ir.model.access.csv
+++ b/hr_attendance_rfid/security/ir.model.access.csv
@@ -1,4 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_hr_attendance_rfid,access.hr.attendance.rfid,hr_attendance.model_hr_attendance,hr_attendance_rfid.group_hr_attendance_rfid,1,1,1,0
-access_hr_employee_rfid,access.hr.employee.rfid,model_hr_employee,hr_attendance_rfid.group_hr_attendance_rfid,1,0,0,0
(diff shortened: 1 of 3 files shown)


hr_attendance_rfid/security/ir.model.access.csv
```diff
@@ -1,4 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_hr_attendance_rfid,access.hr.attendance.rfid,hr_attendance.model_hr_attendance,hr_attendance_rfid.group_hr_attendance_rfid,1,1,1,0
-access_hr_employee_rfid,access.hr.employee.rfid,model_hr_employee,hr_attendance_rfid.group_hr_attendance_rfid,1,0,0,0
```
pedrobaeza, line 3: Why removing this?

victoralmau: Some things about it:

- `model_hr_employee` is not correct (try to install it and will get and error), maybe `hr.model_hr_employee` is correct (xml_id).
- IMO is not needed because now, apply `rfid_card_code` field to `hr.eemployee.base` and with these change, apply to `hr.employee` + `hr.employee.public `and users access to these models acording to access groups (all users access to `hr.employee.public` and users with employee groups, access to hr.employee)

What do you think about it?

pedrobaeza: You still need access to that model for that group, as this is intended for having a user with just that permissions.

pedrobaeza: Maybe you need to grant also access to hr.employee.public to this group?

pedrobaeza (approved):

PR #48, opened 2021-12-02
[FIX] hr_attendance_modification_tracking

brian10048:
The compute method was removed in migration to 14, but the field definition still referenced the compute method


brian10048:
@Saran440 can you please review?

Saran440 (approved):
Code Reviewed :+1:

petrus-v (approved):
LGTM :+1:

PR #50, opened 2021-12-17
[ADD] 14.0 hr_attendance_overtime

petrus-v:
Mark Attendances as overtime according works week employee configuration
to track times done outside the theoretical work times.

At check-out time the attendance line can be split in order to help manager
with employee presence reviews and compute compensatory hours.

More details in description.rst.

Also note that I've developed an other module `hr_attendance_validation` by #53  that will help manager to reveiw hours per weeks and generate compensatory hours from overtime.


hr_attendance_overtime/README.rst
```diff
@@ -0,0 +1,174 @@
...
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/288/14.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Mark Attendances as overtime according works week employee configuration
+to track times done outside the theoretical work times.
+
+At check-out time the attendance line can be split in ordre to help manager
```
ajaniszewska-dev:
```suggestion
At check-out time the attendance line can be split in order to help manager
```

hr_attendance_overtime/README.rst
```diff
@@ -0,0 +1,174 @@
...
+with employee presence reviews and compute compensatoires hours.
```
ajaniszewska-dev:
```suggestion
with employee presence reviews and compute compensatory hours.
```

petrus-v:
@OCA/human-resources-maintainers I've just rebase but it would require someone to explicitly mark as no stale if possible :pray:  ? and probably add the `needs reviews` tags ;)

Thanks deer maintainers !

etobella:
I made a basic test and it seems interesting. However, I miss some constrains and onchange in order to avoid errors when defining all. 

I will make more tests soon I hope :wink:

petrus-v:
> I made a basic test and it seems interesting. However, I miss some constrains and onchange in order to avoid errors when defining all.
> 
> I will make more tests soon I hope wink

Thanks for feedback !

I'm wondering what you mean by "defining all" it this context ? while setting Working Times ?

pedrobaeza:
As information, v15 seems to include an overtime mechanism in core.

petrus-v:
> As information, v15 seems to include an overtime mechanism in core.

Interesting, I haven't notice that !

pedrobaeza:
Check https://github.com/odoo/odoo/pull/69930

nimarosa:
Since odoo v15 includes overtime computation to generate time off credit, it's not the same for cases when you want to track overtime hours like I think this module does.

I didn't tested it but only reading the readme it seems interesting. Anyways, I think that you should resolve the issues you list in "known issues" because I think they are important flaws. 

If you want to finish this and get a stable version that covers the usage you describe, we can merge it. 

I also think you have to add constraints and error handling to make this more usable.

petrus-v:
@nimarosa 

Thanks for review, I'm not sure which issue you refer, probably there is confusion between roadmap.rst which go to the known issue in readme.rst !?

This module is in production for year and can works without those features that I'll give more details bellow. So I don't plan to do them for the time being. Can you precise what to fix or remove the `needs fixing` tag?

## Display planning in kiosk mode as well

In my case we are not using the kiosk mode but as you can see on screen shot, the planing (possible working hours and recorded times) is displayed on user employee screen.

## Ensure it works with a two weeks calendar

I've never use 2 weeks calendar and don't plan to use it, it probably works but as it's not test user should be aware.

## Considering global time off and hr_holidays

The idea behind that is to not be able to check in/check out or switch to overtime while user are on holidays. But this should be definitively done in an other module. Also this is not on the plan anymore, our users don't try to check-in while there are on holidays !

## Ensure there is no overlap between work time lines of the same work time calendar

We do not change a lot work time calendar, we have been rigorous while setting them !
 
 ## Figure out what's wrong with overlap while switching to 2 weeks calendar
 
 As said above I don't use it, if it's a matter to have this issue, I agree to spend time to fix that one. I'll try it to understand I don't remember details about it :sweat_smile:

petrus-v:
> ## Figure out what's wrong with overlap while switching to 2 weeks calendar
> As said above I don't use it, if it's a matter to have this issue, I agree to spend time to fix that one. I'll try it to understand I don't remember details about it sweat_smile

I can't reproduce probably fixed in the meantime, I'll remove this line in ROADMAP.rst...

petrus-v:
hum... I've reproduce with the default `Standard 40 hours/week` but I'm not sure it's related to this module where we get the following message:

> Validation Error
>
> Attendances can't overlap.


I've uninstall this module and got the same message while trying to switch this week config, so definitively not related :)

nimarosa:
@petrus-v yes I'm talking about the two week calendar and overlapping issues that you described in the roadmap. I understand you don't use it but we have to try to provide a stable module in terms of integrating with current hr features.

Like i said I didn't tested it yet so wait for me to test it and then we can continue looking at it. If the issues don't come up, I think it's a good module to have in the repo.

nimarosa (approved):
Okay finally I had time to test it. It works like described and I find it useful module. 

We can continue improving it in OCA thought time, but it works like it says.

nimarosa:
Merging since it tested it extensively since attendance is a module I use a lot. Functional tests look okay.

nimarosa:
Looks like bot is lazy or crashed right now. Will retry tomorrow.

PR #52, opened 2021-12-22
[FIX] hr_attendance_report_theoretical_time: access right to pivot report

yankinmax:
When clicking on the Reporting => Select Employees => View Report stat button of a wizard, a user (that actually has proper rights) might not have enough access right to read the window action.
access-error
Use _for_xml_id for safe openning.


pedrobaeza (approved):

pedrobaeza:
Let's fast-track the obvious patch:

/ocabot merge patch

PR #64, opened 2022-05-19
[ADD] hr_birthday_welcome_message

albariera:
This module adds a birthday message to the welcome message on the employee's birthday.

Captura de pantalla de 2022-05-23 16-00-48


mtelahun (commented):
LGTM code review

However, I don't think your tests are being run. In the test log it fails to load the test because it can't import python module freezegun.

albariera:
You are right, thanks @mtelahun :blush:

hr_birthday_welcome_message/static/src/js/greeting_message.js
```diff
@@ -0,0 +1,30 @@
...
+
+    greeting_message.include({
+        init: function(parent, action) {
+            this._super.apply(this, arguments);
+            this.is_birthday = action.is_birthday;
+        },
+
+        _get_birthday_message: function() {
+            return _t(
+                "HAPPY BIRTHDAY! <i class='fa fa-birthday-cake' /> <br/> Enjoy your day"
```
etobella: Maybe it is not necessary to be upper case, but just my opinion

etobella (approved):
Just a small comment, but LGTM in general

hr_birthday_welcome_message/static/src/js/greeting_message.js
albariera: Changed :)

LoisRForgeFlow (approved):

PR #75, opened 2022-10-18
[14.0][IMP]hr_attendance_reason: turn data into demo data

HekkiMelody:
I'm proposing that the hr.attendance.reason data becomes demo data.

Since the file is noupdate, it wouldn't affect existing installations.


petrus-v (approved):
Good idea !

nimarosa (approved):

nimarosa:
Seems a good idea to me too. Merging. 

/ocabot merge patch

PR #76, opened 2022-10-19
[ADD]  Translation fr.po  hr_attendance_geoocalisation module

rivo2302:


nimarosa:
Hello, translations should be made via Weblate. Check here. 

https://odoo-community.org/page/translate

PR #82, opened 2022-11-10
[15.0][MIG] hr_attendance_user_list: Migration to 15.0

victoralmau:
Migration to 15.0

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT40348


pedrobaeza:
I think this module is not needed existing already "My Attendances" entry. Didn't you see it?

nimarosa:
I think the same of @pedrobaeza , why this module is necessary?

victoralmau:
Ok, sorry, you are right, but what about the pivot view that this module adds?

pedrobaeza:
I don't think it's interesting enough, and if so, it should be another module name like `hr_attendance_pivot` or similar.

victoralmau:
OpenUpgrade: https://github.com/OCA/OpenUpgrade/pull/3616

PR #87, opened 2022-11-21
Update template

kluna1998:
Copier update 14.0

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,9 +1,14 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.6.1
+_commit: v1.11.1
 _src_path: gh:oca/oca-addons-repo-template
 ci: GitHub
 dependency_installation_mode: PIP
 generate_requirements_txt: true
+github_check_license: true
+github_enable_codecov: true
+github_enable_makepot: true
+github_enable_stale_action: true
+github_enforce_dev_status_compatibility: true
 include_wkhtmltopdf: false
 odoo_version: 14.0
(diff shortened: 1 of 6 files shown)


etobella:
@nimarosa @max3903 hr_attendance_sheet tests are failing, but I am unable to know why. Any help?

max3903:
@Urvisha-OSI Can you help us here?

etobella:
As you can see, it already fails on 14.0 base right now. https://github.com/OCA/hr-attendance/actions/runs/3429644672/jobs/5893341628

Maybe something changed on odoo base that makes the changes, I don't know... Any way, some help would be welcome here in order to review the commit I made "fixing" the tests

Urvisha-OSI:
> As you can see, it already fails on 14.0 base right now. https://github.com/OCA/hr-attendance/actions/runs/3429644672/jobs/5893341628
> 
> Maybe something changed on odoo base that makes the changes, I don't know... Any way, some help would be welcome here in order to review the commit I made "fixing" the tests


@max3903 @etobella  all checks are passed already, and I also checked for it in my system there are not any test cases are failed, what do we want more in this PR? @etobella can you please let me know when your test case are fails ?

etobella:
I changed the tests to make it work. Usually, this is the last thing we should do... You can see the error on the previous commits CI or main branch commit.  Can you make sure that the changes applied to the tests are right?

Urvisha-OSI:
> I changed the tests to make it work. Usually, this is the last thing we should do... You can see the error on the previous commits CI or main branch commit. Can you make sure that the changes applied to the tests are right?

yes @etobella, the changes are right, because the old ones are not working

PR #96, opened 2023-01-04
[13.0][IMP] hr_attendance_reason: Show reasons on attendance screen

victoralmau:
Show reasons on attendance screen.
ejemplo

Changes done:
- [x] Add configuration option to show reasons in attendance screen.
- [x] Add field to show reason in attendance screen.
- [x] Add sequence to reasons to allow sorted (from kiosk mode).

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT40897

### DIFF
--- a/hr_attendance_reason/models/hr_attendance_reason.py
+++ b/hr_attendance_reason/models/hr_attendance_reason.py
@@ -8,9 +8,11 @@ from odoo import fields, models
 class HrAttendanceReason(models.Model):
     _name = "hr.attendance.reason"
     _description = "Attendance Reason"
+    _order = "sequence,id"
 
     _sql_constraints = [("unique_code", "UNIQUE(code)", "Code must be unique")]
 
+    sequence = fields.Integer()
     name = fields.Char(
         String="Reason",
         help="Specifies the reason leaving soon or arriving late",
@@ -23,3 +25,4 @@ class HrAttendanceReason(models.Model):
         string="Action Type",
         help="Leave empty if it is independent",
     )
+    show_on_attendance_screen = fields.Boolean(string="Show on attendance screen?")
--- a/hr_attendance_reason/models/res_company.py
+++ b/hr_attendance_reason/models/res_company.py
@@ -1,0 +1,12 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    show_reasons_on_attendance_screen = fields.Boolean(
+        string="Show reasons on attendance screen"
+    )
--- a/hr_attendance_reason/static/src/js/kiosk_confirm.js
+++ b/hr_attendance_reason/static/src/js/kiosk_confirm.js
@@ -1,0 +1,57 @@
+odoo.define("hr_attendance_reason.kiosk_confirm", function(require) {
+    "use strict";
+    const KioskConfirm = require("hr_attendance.kiosk_confirm");
+    const session = require("web.session");
+    const core = require("web.core");
+    const _t = core._t;
+    const superSignInOut =
+        KioskConfirm.prototype.events["click .o_hr_attendance_sign_in_out_icon"];
+    const superPinPadButton =
+        KioskConfirm.prototype.events["click .o_hr_attendance_pin_pad_button_ok"];
+    KioskConfirm.include({
+        events: Object.assign({}, KioskConfirm.prototype.events, {
+            "click .o_hr_attendance_sign_in_out_icon": function() {
+                this.update_attendance(superSignInOut);
+            },
+            "click .o_hr_attendance_pin_pad_button_ok": function() {
+                this.update_attendance(superPinPadButton);
+            },
+        }),
+        willStart: async function() {
+            await this._super();
+            await this._rpc({
+                model: "hr.employee",
+                method: "search_read",
+                args: [
+                    [["user_id", "=", this.getSession().uid]],
+                    ["show_reasons_on_attendance_screen"],
+                ],
+            }).then(res => {
+                this.employee = res.length && res[0];
+            });
+            await this._rpc({
+                model: "hr.attendance.reason",
+                method: "search_read",
+                fields: ["name", "action_type"],
+                domain: [["show_on_attendance_screen", "=", true]],
+            }).then(reasons => {
+                this.reasons = reasons;
+            });
+        },
+        update_attendance: function(event_func) {
+            const attendance_reason_id = parseInt(
+                this.$(".o_hr_attendance_reason").val(),
+                0
+            );
+            Object.assign(session.user_context, {
+                attendance_reason_id: attendance_reason_id,
+            });
+            if (attendance_reason_id === 0) {
+                this.do_warn(_t("Please, select a reason"));
+            } else {
+                event_func = event_func.bind(this);
+                event_func();
+            }
+        },
+    });
+});
--- a/hr_attendance_reason/static/src/js/my_attendances.js
+++ b/hr_attendance_reason/static/src/js/my_attendances.js
@@ -1,0 +1,47 @@
+odoo.define("hr_attendance_reason.my_attendances", function(require) {
+    "use strict";
+
+    var MyAttendances = require("hr_attendance.my_attendances");
+    const session = require("web.session");
+    var core = require("web.core");
+    var _t = core._t;
+
+    MyAttendances.include({
+        willStart: function() {
+            Object.assign(session.user_context, {
+                extra_fields: ["show_reasons_on_attendance_screen"],
+            });
+            return this._super();
+        },
+        init: function() {
+            this._super.apply(this, arguments);
+            this.reasons = this.getAttendanceReasons();
+        },
+        getAttendanceReasons: function() {
+            this.reasons = [];
+            this._rpc({
+                model: "hr.attendance.reason",
+                method: "search_read",
+                fields: ["name", "action_type"],
+                domain: [["show_on_attendance_screen", "=", true]],
+            }).then(reasons => {
+                this.reasons = reasons;
+            });
+            return this.reasons;
+        },
+        update_attendance: function() {
+            const attendance_reason_id = parseInt(
+                this.$(".o_hr_attendance_reason").val(),
+                0
+            );
+            Object.assign(session.user_context, {
+                attendance_reason_id: attendance_reason_id,
+            });
+            if (attendance_reason_id === 0) {
+                this.do_warn(_t("Please, select a reason"));
+            } else {
+                this._super();
+            }
+        },
+    });
+});
--- a/hr_attendance_reason/static/src/xml/attendance.xml
+++ b/hr_attendance_reason/static/src/xml/attendance.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates>
+    <t t-extend="HrAttendanceMyMainMenu">
+        <t t-jquery="h4:first" t-operation="after">
+            <t t-if="widget.employee.show_reasons_on_attendance_screen">
+                <div class="reasons_lists" t-if="widget.reasons">
+                    <h3 class="mt8">Reason</h3>
+                    <select
+                        class="o_hr_attendance_reason o_input_dropdown o_input o_field_widget o_required_modifier"
+                    >
+                        <option value="0" />
+                        <t t-foreach="widget.reasons" t-as="reason">
+                            <option
+                                t-att-value="reason.id"
+                                t-esc="reason.name"
+                                t-if="reason.action_type === 'sign_in' and !checked_in"
+                            />
+                            <option
+                                t-att-value="reason.id"
+                                t-esc="reason.name"
+                                t-if="reason.action_type === 'sign_out' and checked_in"
+                            />
+                        </t>
+                    </select>
+                </div>
+            </t>
+        </t>
+    </t>
+    <t t-extend="HrAttendanceKioskConfirm">
+        <t t-jquery="h4:first" t-operation="after">
+            <t t-if="widget.employee.show_reasons_on_attendance_screen">
+                <div class="reasons_lists" t-if="widget.reasons">
+                    <h3 class="mt8">Reason</h3>
+                    <select
+                        class="o_hr_attendance_reason o_input_dropdown o_input o_field_widget o_required_modifier"
+                    >
+                        <option value="0" />
+                        <t t-foreach="widget.reasons" t-as="reason">
+                            <option
+                                t-att-value="reason.id"
+                                t-esc="reason.name"
+                                t-if="reason.action_type === 'sign_in' and !checked_in"
+                            />
+                            <option
+                                t-att-value="reason.id"
+                                t-esc="reason.name"
+                                t-if="reason.action_type === 'sign_out' and checked_in"
+                            />
+                        </t>
+                    </select>
+                </div>
+            </t>
+        </t>
+    </t>
+</templates>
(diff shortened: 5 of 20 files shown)


hr_attendance_reason/models/hr_attendance_reason.py
```diff
@@ -8,9 +8,11 @@
 class HrAttendanceReason(models.Model):
     _name = "hr.attendance.reason"
     _description = "Attendance Reason"
+    _order = "sequence"
```
pedrobaeza: Remember always to put id as fallback.

```suggestion
    _order = "sequence,id"
```

hr_attendance_reason/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    show_in_kiosk_mode = fields.Boolean(string="Show reasons")
```
pedrobaeza: As this is a general model, you need to be explicit in the name:

```suggestion
    attendance_reasons_in_kiosk = fields.Boolean(string="Show attendance reasons in Kiosk")
```

pedrobaeza (changes requested):
- You are not adding the reason to the kiosk mode, but the regular attendance screen.
- The shown reasons appear randomly. Sometimes "Felt sick", sometimes "Visit customer"...
- No default reason should be selected, and you can't click on check-in if no reason is selected.

nimarosa (changes requested):
Apart of the comments of @pedrobaeza  which I agree on I suggest:

- Use some odoo css clases to format the drop-down so it matches with the UI
- Only let the user choose reasons if the time of checking out/checking in is out of the employee schedule. If the employee it's checking in/out at the right time I suggest not to show the drop-down because it could be confusing. 
- Do not provide a default value, it can cause bad inputs.

victoralmau:
> * You are not adding the reason to the kiosk mode, but the regular attendance screen.

Ok, thank you for the clarification.

> * The shown reasons appear randomly. Sometimes "Felt sick", sometimes "Visit customer"...

They are not displayed randomly, as defined, some reasons will be displayed for check_in and different reasons for check_out.
(perhaps the existing reasons are somewhat unclear, I agree).

> * No default reason should be selected, and you can't click on check-in if no reason is selected.

Ok, changes done. Now if a reason is not selected, an error is displayed.

victoralmau:
> * Use some odoo css clases to format the drop-down so it matches with the UI

Changes done.

> * Only let the user choose reasons if the time of checking out/checking in is out of the employee schedule. If the employee it's checking in/out at the right time I suggest not to show the drop-down because it could be confusing.

While perhaps a good suggestion, I think this is beyond the scope of this enhancement; moreover, it would be necessary to add a screen refresh every x minutes to "know" whether or not to display the reasons.

In addition, in some cases you may want to allow the reason for check out (e.g. for the coffee break).

> * Do not provide a default value, it can cause bad inputs.

Changes done.

pedrobaeza (changes requested):
Remove the kiosk word from everywhere, as it's not being applied to kiosk mode. And I think it should work also on kiosk mode. I would call it "Show reasons on attendance screen".

hr_attendance_reason/static/src/xml/attendance.xml
```diff
@@ -0,0 +1,29 @@
...
+                    <h3 class="mt8">Reason</h3>
+                    <select
+                        class="o_hr_attendance_reason o_input_dropdown o_input o_field_widget o_required_modifier"
+                    >
+                        <option value="0" />
+                        <t t-foreach="widget.reasons" t-as="reason">
+                            <option
+                                t-att-value="reason.id"
+                                t-esc="reason.name"
+                                t-if="reason.action_type == 'sign_in' and checked_in"
```
chienandalu: Wouldn't it be the opposite?
```suggestion
                                t-if="reason.action_type === 'sign_in' and !checked_in"
```

hr_attendance_reason/static/src/xml/attendance.xml
```diff
@@ -0,0 +1,29 @@
...
+                            />
+                            <option
+                                t-att-value="reason.id"
+                                t-esc="reason.name"
+                                t-if="reason.action_type == 'sign_out' and !checked_in"
```
chienandalu: And the same here
```suggestion
                                t-if="reason.action_type === 'sign_out' and checked_in"
```

hr_attendance_reason/static/src/js/attendance_reason.js
```diff
@@ -0,0 +1,69 @@
...
+                fields: ["name", "action_type"],
+                domain: [["show_in_kiosk_mode", "=", true]],
+            }).then(function(reasons) {
+                _.each(reasons, function(reason) {
+                    self.reasons.push(reason);
+                });
+            });
+            return this.reasons;
+        },
+        update_attendance: function() {
```
chienandalu: Isn't this shadowing the original function?

hr_attendance_reason/static/src/xml/attendance.xml
```diff
@@ -0,0 +1,29 @@
...
+    <t t-extend="HrAttendanceMyMainMenu">
+        <t t-jquery="h4:first" t-operation="after">
+            <t t-if="widget.attendance_reasons_in_kiosk">
+                <div class="reasons_lists" t-if="widget.reasons">
+                    <h3 class="mt8">Reason</h3>
+                    <select
+                        class="o_hr_attendance_reason o_input_dropdown o_input o_field_widget o_required_modifier"
+                    >
+                        <option value="0" />
+                        <t t-foreach="widget.reasons" t-as="reason">
```
chienandalu, line 12: Although maybe you can do it in one go here. Somethin like this (not tested):
```suggestion
                        <t t-foreach="widget.reasons.filter(reason => {return (checked_in && reason.action_type === 'sign_out') || (!checked_in && reason.action_type === 'sign_in')})" t-as="reason">
```

chienandalu (changes requested):

hr_attendance_reason/static/src/xml/attendance.xml
victoralmau: This and similar things don't work, I don't know exactly the syntax to use in this use case, sorry.

hr_attendance_reason/static/src/js/attendance_reason.js
victoralmau: Yes, the whole function is replaced. Reviewing the code of `hr_attendance` I see that it is a "hell" inheritance (not possible) and it is necessary to re-define some methods completely again. Any suggestions on how to simplify the code (js especially)?

victoralmau:
> Remove the kiosk word from everywhere, as it's not being applied to kiosk mode. And I think it should work also on kiosk mode. I would call it "Show reasons on attendance screen".

Removed the word kiosk completely and added functionality to kiosk mode.

hr_attendance_reason/static/src/xml/attendance.xml
chienandalu: Looks like the compiler messes the lambda arguments up. Not much to worry about anyway :)

hr_attendance_reason/static/src/js/attendance_reason.js
chienandalu: Not easy to solve and we'll clash with this other module which does the same monkeypatching https://github.com/OCA/hr-attendance/blob/cc0234df35d7fda1d090c121f0799d5c7fc70e72/hr_attendance_geolocation/static/src/js/attendance_geolocation.js#L14-L28

pedrobaeza: We need then a base module that do this monkey-patching once and allow inheritance in dependent modules.

hr_attendance_reason/static/src/js/attendance_reason.js
```diff
@@ -0,0 +1,162 @@
...
+                    context: context,
+                }).then(function(result) {
+                    if (result.action) {
+                        self.do_action(result.action);
+                    } else if (result.warning) {
+                        self.do_warn(result.warning);
+                    }
+                });
+            }
+        },
```
chienandalu: In this case we can use `session.user_context` to respect inheritance:
```suggestion
            const attendance_reason_id = parseInt(
                this.$(".o_hr_attendance_reason").val(),
                0
            );
            Object.assign(
                session.user_context,
                {attendance_reason_id: attendance_reason_id}
            )
            if (attendance_reason_id === 0) {
                this.do_warn(_t("Please, select a reason"));
            } else {
                this._super();
            }
        },
```

hr_attendance_reason/static/src/js/attendance_reason.js
chienandalu: For this case I researched a solution that fits the case. For the other module, I should find out.

pedrobaeza: OK, let's use that method then for now.

hr_attendance_reason/static/src/js/attendance_reason.js
```diff
@@ -0,0 +1,162 @@
...
+                fields: ["name", "action_type"],
+                domain: [["show_on_attendance_screen", "=", true]],
+            }).then(function(reasons) {
+                _.each(reasons, function(reason) {
+                    self.reasons.push(reason);
+                });
+            });
+            return this.reasons;
+        },
+        update_attendance: function() {
```
chienandalu: This could be inherited as suggested in the other widget with `session.user_context`

hr_attendance_reason/static/src/js/kiosk_confirm.js
```diff
@@ -0,0 +1,91 @@
...
+                            self.$(".o_hr_attendance_pin_pad_button_ok").removeAttr(
+                                "disabled"
+                            );
+                        }, 500);
+                    }
+                });
+            }
+        },
+    });
+});
```
chienandalu: We can wrap the event functions allowing us to do the right patching:
```suggestion
odoo.define("hr_attendance_reason.kiosk_confirm", function(require) {
    "use strict";

    const KioskConfirm = require("hr_attendance.kiosk_confirm");
    const session = require("web.session");
    const core = require("web.core");
    const _t = core._t;

    const superSignInOut = KioskConfirm.prototype.events["click .o_hr_attendance_sign_in_out_icon"];
    const superPinPadButton = KioskConfirm.prototype.events["click .o_hr_attendance_pin_pad_button_ok"];

    KioskConfirm.include({
        events: Object.assign({}, KioskConfirm.prototype.events, {
            "click .o_hr_attendance_sign_in_out_icon": function() {
                this.update_attendance(superSignInOut);
            },
            "click .o_hr_attendance_pin_pad_button_ok": function () {
                this.update_attendance(superPinPadButton);
            },
        }),
        willStart: async function() {
            await this._super();
            await this._rpc({
                model: "hr.employee",
                method: "search_read",
                args: [
                    [["user_id", "=", this.getSession().uid]],
                    ["show_reasons_on_attendance_screen"],
                ],
            }).then((res) => {
                this.employee = res.length && res[0];
            });
            await this._rpc({
                model: "hr.attendance.reason",
                method: "search_read",
                fields: ["name", "action_type"],
                domain: [["show_on_attendance_screen", "=", true]],
            }).then((reasons) => {
                this.reasons = reasons;
            });
        },
        update_attendance: function(event_func) {
            const attendance_reason_id = parseInt(
                this.$(".o_hr_attendance_reason").val(),
                0
            );
            Object.assign(session.user_context, {
                attendance_reason_id: attendance_reason_id,
            });
            if (attendance_reason_id === 0) {
                this.do_warn(_t("Please, select a reason"));
            } else {
                event_func = event_func.bind(this);
                event_func();
            }
        },
    });
});
```

victoralmau:
Thanks David, now everything works correctly.

hr_attendance_reason/static/src/js/my_attendances.js
```diff
@@ -0,0 +1,51 @@
...
+    MyAttendances.include({
+        willStart: function() {
+            Object.assign(session.user_context, {
+                extra_fields: ["show_reasons_on_attendance_screen"],
+            });
+            return this._super();
+        },
+        // eslint-disable-next-line no-unused-vars
+        init: function(parent, action) {
+            this._super.apply(this, arguments);
```
chienandalu: It'll be the same:
```suggestion
        init: function() {
            this._super.apply(this, arguments);
```

hr_attendance_reason/static/src/js/my_attendances.js
```diff
@@ -0,0 +1,51 @@
...
+            this.reasons = this.getAttendanceReasons();
+        },
+        getAttendanceReasons: function() {
+            var self = this;
```
chienandalu:
```suggestion
```

hr_attendance_reason/static/src/js/my_attendances.js
```diff
@@ -0,0 +1,51 @@
...
+            this.reasons = [];
+            this._rpc({
+                model: "hr.attendance.reason",
+                method: "search_read",
+                fields: ["name", "action_type"],
+                domain: [["show_on_attendance_screen", "=", true]],
+            }).then(function(reasons) {
+                _.each(reasons, function(reason) {
+                    self.reasons.push(reason);
```
chienandalu: Simpler:
```suggestion
            }).then((reasons) => {
                this.reasons = reasons;
```

chienandalu (approved):
:smile: :+1: Some minor extra comments (non blocking):

chienandalu:
@nimarosa Can you update your review?

pedrobaeza:
Got this error checking-in:

```
Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 624, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/opt/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/opt/odoo/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/addons/web/controllers/main.py", line 1342, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/main.py", line 1334, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 390, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 377, in _call_kw_multi
    result = method(recs, *args, **kwargs)
TypeError: attendance_manual() takes from 2 to 3 positional arguments but 4 were given
```

victoralmau:
Rebase done and fixed now (tested in runboat)

nimarosa (approved):
Code review looks good now for me. 
No time to test it functional but looks good. Will try in runboat later tomorrow. 

Are you going to port this to 14.0 15.0 and 16.0 branches too? 

Thanks

victoralmau:
Any other functional review to merge it?

pedrobaeza (approved):
Please promote Tecnativa as co-author, due to the big changes that modify totally the module.

victoralmau:
> Please promote Tecnativa as co-author, due to the big changes that modify totally the module.

Added.

PR #80, opened 2022-11-04
[15.0][MIG] hr_attendance_geolocation: Migration to 15.0

victoralmau:
Migration to 15.0

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT40358

PR #98, opened 2023-01-05
[15.0][IMP] hr_attendance_geolocation: Change js to prevent full override attendance_manual() method.

victoralmau:
Change js to prevent full override `attendance_manual()` method.

TODO: The code should be improved to avoid having a private method (_manual_attendance) instead of calling the super (update_attendance) and to allow compatibility with other modules.

Please @chienandalu can you review it?

@Tecnativa

### DIFF
--- a/hr_attendance_geolocation/static/src/js/attendance_geolocation.js
+++ b/hr_attendance_geolocation/static/src/js/attendance_geolocation.js
@@ -29,15 +29,18 @@ odoo.define("hr_attendance_geolocation.attendances_geolocation", function (requi
         },
         _manual_attendance: function (position) {
             var self = this;
+            const ctx = Object.assign(session.user_context, {
+                latitude: position.coords.latitude,
+                longitude: position.coords.longitude,
+            });
             this._rpc({
                 model: "hr.employee",
                 method: "attendance_manual",
                 args: [
                     [self.employee.id],
                     "hr_attendance.hr_attendance_action_my_attendances",
-                    null,
-                    [position.coords.latitude, position.coords.longitude],
                 ],
+                context: ctx,
             }).then(function (result) {
                 if (result.action) {
                     self.do_action(result.action);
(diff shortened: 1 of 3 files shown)


hr_attendance_geolocation/static/src/js/attendance_geolocation.js
```diff
@@ -29,15 +29,18 @@ odoo.define("hr_attendance_geolocation.attendances_geolocation", function (requi
         },
         _manual_attendance: function (position) {
             var self = this;
+            var ctx = _.extend({}, session.user_context, {
+                latitude: position.coords.latitude,
+                longitude: position.coords.longitude,
+            });
```
chienandalu:
```suggestion
            var const = Object.assign(session.user_context, {
                latitude: position.coords.latitude,
                longitude: position.coords.longitude,
            });
```

hr_attendance_geolocation/static/src/js/attendance_geolocation.js
```diff
@@ -106,16 +109,15 @@ odoo.define("hr_attendance_geolocation.attendances_geolocation", function (requi
                 );
                 pinBoxVal = this.$(".o_hr_attendance_PINbox").val();
             }
+            var ctx = _.extend({}, session.user_context, {
+                latitude: position.coords.latitude,
+                longitude: position.coords.longitude,
+            });
```
chienandalu: Better to go with vanilla JS and avoid underscore as much as possible
```suggestion
            const ctx = Object.assign(session.user_context, {
                latitude: position.coords.latitude,
                longitude: position.coords.longitude,
            });
```

hr_attendance_geolocation/tests/test_hr_attendance_geolocation.py
```diff
@@ -14,9 +15,9 @@ def setUp(self):
 
     def test_attendance_geolocation(self):
         # Called from js
-        self.employee.attendance_manual(
-            "hr_attendance.hr_attendance_action_my_attendances", None, self.location
-        )
+        self.employee.with_context(
+            latitude=self.location[0], longitude=self.location[1]
+        ).attendance_manual({})
```
chienandalu:
```suggestion
        lat, long = self.location
        self.employee.with_context(
            latitude=lat, longitude=long
        ).attendance_manual({})
```

chienandalu (approved):
I tried to test it on runbot but it seems that the browser considers it non secure so it won't gather the geolocation.

Non blocking comments:

victoralmau:
> I tried to test it on runbot but it seems that the browser considers it non secure so it won't gather the geolocation.

You are right, it cannot currently be tested on runboat because https is not available. (related to https://github.com/OCA/hr-attendance/pull/80#issuecomment-1308693121)

> Non blocking comments:

Thanks, changes done.

chienandalu:
Ok, then :)

pedrobaeza:
Just change `http` to `https` on the URL and you will be able to test it accepting the warning.

chienandalu:
Ok, thanks. Tested then :smile:

pedrobaeza (approved):

PR #99, opened 2023-01-09
[13.0][FIX] hr_attendance_geolocation: inheritance

chienandalu:
All the JS code has been refactored to use context from js and avoid agressive patching that disallows correct inheritance.

Please review @pedrobaeza @victoralmau 

Related work: https://github.com/OCA/hr-attendance/pull/96

You're opinion as author is very welcome as well @AdriaGForgeFlow

### DIFF
--- a/hr_attendance_geolocation/README.rst
+++ b/hr_attendance_geolocation/README.rst
@@ -50,11 +50,15 @@ Authors
 ~~~~~~~
 
 * ForgeFlow S.L.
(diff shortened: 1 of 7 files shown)


victoralmau (approved):
Very good work, code and functional review OK.

pedrobaeza (approved):
/ocabot merge major

Please fw-port it to upper versions

PR #103, opened 2023-01-18
[13.0][IMP] hr_attendance_report_theoretical_time: Allow inheritance

victoralmau:
Allow inheritance

@Tecnativa TT41242


victoralmau:
Finally, it is not necessary, it can be extended with existing methods.

PR #100, opened 2023-01-11
[14.0][FIX] hr_attendance_geolocation: inheritance

chienandalu:
FW of #99 

Use session context to inject position avoiding agressive patching

cc @Tecnativa

please review @pedrobaeza @victoralmau

PR #104, opened 2023-01-23
[15.0][IMP] hr_attendance_geolocation: Refactor javascript

Tardo:
More simple, less invasive...

Before this PR it was not possible to use this module together with other modules that modify the "kiosk" view. After this PR, this module is compatible with any other module that touches "hr attendance".


pedrobaeza:
Please fw-port #100 instead.

Tardo:
Direct fw-port is not possible due to restrictions in 15.0+ ... you can't modify directly the user_context in "legacy env": https://github.com/odoo/odoo/blob/15.0/addons/web/static/src/legacy/utils.js#L120

For this need something like:  https://github.com/OCA/hr-attendance/pull/104/files#diff-6d083333fcf54892937188885c2cb858fd10a884790ef7be8b5139df21cf740aR12

Tardo:
Changes applied. I hope everything is ok now.

Tests errors comes from "hr_attendance_report_theoretical_time" module.

hr_attendance_geolocation/static/src/js/geolocation_user_context_legacy.esm.js
```diff
@@ -0,0 +1,34 @@
...
+                const nvals = {};
+                if ("longitude" in values) {
+                    nvals.longitude = values.longitude;
+                }
+                if ("latitude" in values) {
+                    nvals.latitude = values.latitude;
+                }
+                if (!_.isEmpty(nvals)) {
+                    env.services.user.updateContext(nvals);
+                }
```
chienandalu: We can simplify it using destructuring:
```suggestion
                const {latitude, longitude} = vals;
                if (latitude && longitude) {
                    env.services.user.updateContext({
                        latitude: latitude,
                        longitude: longitude,
                    });
                }
```

hr_attendance_geolocation/tests/test_hr_attendance_geolocation.py
```diff
@@ -15,8 +15,9 @@ def setUp(self):
 
     def test_attendance_geolocation(self):
         # Called from js
-        lat, long = self.location
-        self.employee.with_context(latitude=lat, longitude=long).attendance_manual({})
+        self.employee.with_context(attendance_location=self.location).attendance_manual(
+            "hr_attendance.hr_attendance_action_my_attendances"
+        )
```
chienandalu: This should stay as it is in v15
```suggestion
        lat, long = self.location
        self.employee.with_context(latitude=lat, longitude=long).attendance_manual({})
```

hr_attendance_geolocation/static/src/js/geolocation_service_legacy.js
```diff
@@ -0,0 +1,56 @@
...
+
+    const core = require("web.core");
+    const AbstractService = require("web.AbstractService");
+    const session = require("web.session");
+
+    const GeolocationService = AbstractService.extend({
+        dependencies: [],
+
+        start: function () {
+            this._hasGeolocation = typeof navigator.geolocation !== "undefined";
```
chienandalu, line 15: Wouldn't it be?:
```suggestion
            this._hasGeolocation = typeof navigator.geolocation !== undefined;
```

hr_attendance_geolocation/static/src/js/attendance_geolocation.js
```diff
@@ -1,149 +1,32 @@
...
-                },
-            };
-            this._manual_attendance(position);
+            this.call("geolocation_service", "updateUserContext");
         },
     });
 
+    // With Kiosk, we had no method to override, as the event property that triggers the
+    // update is mapped directly to a function on the fly. In this case, we'll use the
+    // asyncronous willStart method so we can wait for the location to be resolved.
```
chienandalu, line 28: This comment isn't true anymore

chienandalu (changes requested):
Hi, Alex! Good to see you around here :D

Thanks for your nice solution! I got stuck on the context getter but I couldn't get around it. I proposed a fix to Odoo https://github.com/odoo/odoo/pull/109822 but there's no response yet. Maybe we can put a comment to simplify the implementation if it's accepted someday...

Some comments below:

hr_attendance_geolocation/static/src/js/geolocation_user_context_legacy.esm.js
Tardo: We don't currently use it... but if we make this change, the values cannot be updated individually.

hr_attendance_geolocation/static/src/js/geolocation_service_legacy.js
Tardo: typeof returns a string: https://developer.mozilla.org/es/docs/Web/JavaScript/Reference/Operators/typeof

hr_attendance_geolocation/static/src/js/geolocation_user_context_legacy.esm.js
chienandalu: Why would you?

Tardo:
@chienandalu changes done... and now still working in async mode.

hr_attendance_geolocation/static/src/js/geolocation_user_context_legacy.esm.js
```diff
@@ -0,0 +1,31 @@
...
+import * as legacySession from "web.session";
+
+const LegacyGelocationSessionUserContextService = {
+    dependencies: ["user"],
+
+    start(env) {
+        Object.defineProperty(legacySession, "user_geolocation_context", {
+            set: (values) => {
+                const {latitude, longitude} = values;
+                if (typeof latitude !== 'undefined' && typeof longitude !== 'undefined') {
```
chienandalu: Why not reliying on type coercion?

Tardo: Because 0 (zero) should be a valid value... in case of error the value is setted to 0.

chienandalu: Sure :sweat_smile: :+1:

chienandalu (approved):
Thanks!

chienandalu:
I'm going to try to fix CI tests here: https://github.com/OCA/hr-attendance/pull/105

The issue comes from this commit which asumes that the user tz is always set https://github.com/odoo/odoo/commit/d6f2437e9ce69a585c33940c6a2c18f4a6c7fdad#diff-38469def2f870bb866f971f57797dd7c21b6a95d52a8eae72f832f0eea2434f9R105

chienandalu:
You can rebase @Tardo

hr_attendance_geolocation/README.rst
```diff
@@ -13,15 +13,15 @@ Hr Attendance Geolocation
 .. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
-.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fhr--attendance-lightgray.png?logo=github
-    :target: https://github.com/OCA/hr-attendance/tree/15.0/hr_attendance_geolocation
-    :alt: OCA/hr-attendance
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fhr_attendance-lightgray.png?logo=github
+    :target: https://github.com/OCA/hr_attendance/tree/15.0/hr_attendance_geolocation
```
pedrobaeza: This change is not correct, as the repository is `hr-attendance`, not `hr_attendance`.

Tardo: Oops! updated

pedrobaeza (approved):
/ocabot merge patch

PR #105, opened 2023-01-24
[15.0][FIX] hr_attendance_report_theoretical_time: tests

chienandalu:
Tests rely on removing tz from the current user and it isn't possible anymore after https://github.com/odoo/odoo/commit/d6f2437e9ce69a585c33940c6a2c18f4a6c7fdad#diff-38469def2f870bb866f971f57797dd7c21b6a95d52a8eae72f832f0eea2434f9R105

### DIFF
--- a/hr_attendance_report_theoretical_time/tests/test_hr_attendance_report_theoretical_time.py
+++ b/hr_attendance_report_theoretical_time/tests/test_hr_attendance_report_theoretical_time.py
@@ -98,8 +98,6 @@ class TestHrAttendanceReportTheoreticalTimeBase(common.TransactionCase):
                 "requires_allocation": "no",
             }
         )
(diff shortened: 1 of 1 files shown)


pedrobaeza:
That was put for test resilience, so the reason why it was put will arise again on one point. Do you know why it was put using git blame?

chienandalu:
It goes all the way back here but I couldn't find out the reason why it was needed originally: https://github.com/OCA/hr/pull/465/commits/7c38f50973b0dbd0ee3de49ad39792535dafa032

Tardo:
Maybe someone put it in thinking that changing to False would force times to be displayed in UTC?

Tardo (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #110, opened 2023-02-16
[FIX] hr_attendance_modification_tracking: Set related fields as not tracking

etobella:
Without this change, all the check_in and check_out are displayed on the employee

image


sanchonuria (approved):

kluna1998 (approved):

pedrobaeza:
Is this a side effect of the module?

etobella:
Yep, the tracking was intended for hr.attendance, but as it is using a related field, the tracking is inherited there. We need to enforce that it is not tracked on hr.employee.base :cry:

PR #113, opened 2023-03-28
[14.0][FIX] hr_attendance_modification_tracking: squash tracking warning

HekkiMelody:
Removes the warning in the logs about "tracking" not being allowed on `hr.employee.public`

```
odoo.fields: Field hr.employee.public.last_check_in: unknown parameter 'tracking', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it 
odoo.fields: Field hr.employee.public.last_check_out: unknown parameter 'tracking', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it 
```

Ref #110

### DIFF
--- a/hr_attendance_modification_tracking/models/hr_employee.py
+++ b/hr_attendance_modification_tracking/models/hr_employee.py
@@ -9,3 +9,6 @@ class HrEmployeeBase(models.AbstractModel):
 
     last_check_in = fields.Datetime(tracking=False)
     last_check_out = fields.Datetime(tracking=False)
+
+    def _valid_field_parameter(self, field, name):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why not doing it on the base abstract model?

HekkiMelody:
> Why not doing it on the base abstract model?

I don't mind doing it that way. The warning only shows on that specific model (because it doesn't inherit from mail.thread) so I only overwrote it there.

I'll change it.

pedrobaeza (approved):

nimarosa (approved):

PR #114, opened 2023-03-28
[15.0][MIG] hr_attendance_sheet: Migration to 15.0

albig:


max3903:
@sbidoul I am not sure why but I don't have the permissions to set labels, milestones or reviewers or use ocabot in this repo. Can you help?

max3903:
@albig Can you add more tests to increase code coverage? Thank you

sbidoul:
>  I am not sure why but I don't have the permissions to set labels, milestones or reviewers or use ocabot in this repo. Can you help?

Hi @max3903, it seems you are not part of the [Human Resources PSC](https://odoo-community.org/psc-teams/human-resources-19), this explains that.

max3903:
@dreispt Can you help?
@patrickrwilson Are you interested to join the HR PSC?
@albig I don't mind being removed as a maintainer of this module.

Tardo:
I have a few questions:

- Why are the theoretical hours not computed against the employee's attendance calendar?
- Why is the hr.attendance.overtime model not used to calculate the overtime?

albig:
Thanks @Tardo for your good questions. 

I don't have the answers, because I haven't developped this module.

> * Why are the theoretical hours not computed against the employee's attendance calendar?

That's what I expected, too. This implementation is too basic but maybe I don't get the use case.

> * Why is the hr.attendance.overtime model not used to calculate the overtime?

For sure, because hr.attendance.overtime was introduced with Odoo 15 and this module arrived in this OCA repository in version 14 (maybe much older?).

Tardo:
> That's what I expected, too. This implementation is too basic but maybe I don't get the use case.

I would say that it is done this way in order to be able to create sheets with different calendars (hours). I think this should be handled by another module that allows you to define attendance calendars by dates.

This is an important change and depends on this new module. I don't know what people think about making this effort... or if any changes are open for discussion.

> For sure, because hr.attendance.overtime was introduced with Odoo 15 and this module arrived in this OCA repository in version 14 (maybe much older?).

I would say that an effort should be made to switch to using this new model.


**Perhaps, as they are changes in the operation of the module, it would be interesting to make a new module that changes the operation of this one and leave this one as it is.**

Related module: hr_contract_employee_calendar_planning

albig:
We do not follow this approach anymore.

PR #118, opened 2023-04-06
[16.0] [mig] hr_attendance_modification_tracking

aktiv-heli-kantawala:


hr_attendance_modification_tracking/models/hr_attendance.py
```diff
@@ -0,0 +1,55 @@
...
+        now = fields.Datetime.now()
+        for val in vals:
+            for check in ["check_in", "check_out"]:
+                if (
+                    val.get(check, False)
+                    and abs(fields.Datetime.from_string(val.get(check)) - now)
+                    > tolerance
+                ):
+                    val.update({"time_changed_manually": True})
+                    break
```
HekkiMelody, line 38: I'd call the initial `vals` --> `vals_list` and then what you called `val` is actually `vals` or `vals_dict`

HekkiMelody (commented):
Code review, mostly ok.

The last commit with the pre-commit execution shouldn't be squashed, as there is a previous commit also about pre-commit execution. Aside from that one (whose message starts with ^C) pre-commit should be run always.

hr_attendance_modification_tracking/models/hr_attendance.py
```diff
@@ -0,0 +1,55 @@
...
+        tolerance = timedelta(seconds=60)
+        now = fields.Datetime.now()
+        for record in self:
+            for check in ["check_in", "check_out"]:
+                if vals.get(check, False):
+                    if getattr(record, check, False):
+                        record.time_changed_manually = True
+                    else:
+                        check_str = vals.get(check)
+                        diff = abs(fields.Datetime.from_string(check_str) - now)
```
dalonsofl, line 51: Maybe it's not necessary the second condition (
`

        if vals.get(check, False):
            if getattr(record, check, False):
                record.time_changed_manually = True
            else:
                check_str = vals.get(check)
`
) , you can try use the first if as a variable:

`


         check_val = vals.get(check, False)
                if check_val:
                    check_time = fields.Datetime.from_string(check_val)
                    diff = abs(check_time - now)`

hr_attendance_modification_tracking/models/hr_attendance.py
```diff
@@ -0,0 +1,55 @@
...
+                        if diff > tolerance:
+                            record.time_changed_manually = True
+
+        return super().write(vals)
```
dalonsofl, line 55: And, at the end, i will check if any `record.time_changed_manually` is True, so write the vals:

`

     if any(record.time_changed_manually for record in self):
              return super().write(vals)
     return True
`

hr_attendance_modification_tracking/tests/test_hr_attendance.py
```diff
@@ -0,0 +1,122 @@
...
+from odoo.tests.common import TransactionCase
+from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT as DF
+
+
+class TestHrAttendanceTracking(TransactionCase):
+    def setUp(self):
+        super(TestHrAttendanceTracking, self).setUp()
+        self.hr_attendance = self.env["hr.attendance"]
+
+    def test_attendance_edit(self):
```
dalonsofl, line 18: I think it would be better to split the tests into some tests in order to focus on single cases than all cases in one test.

kikopeiro:
@heliaktiv Could you review the comments? Thanks!

flalexg (approved):
LGTM!

HekkiMelody:
bump

pedrobaeza:
Superseded by #145

PR #121, opened 2023-04-13
[16.0][MIG] hr_attendance_report_theoretical_time: Migration to 16.0

dzungtran89:
While migrating the module, I'm quite confused about the flow of some code parts, so I just would like to understand them

- At this [code part](https://github.com/OCA/hr-attendance/blob/15.0/hr_attendance_report_theoretical_time/models/hr_leave.py#L32), the hours `record.date_from` and `record.date_to` are replaced to 0 and 23 respectively

  - Assume that my timezone is GMT+7
  - I have a leave on 2023-04-10, time is from 8am to 5pm, then UTC time converted will be 1am and 10am respectively
  - Then I have an attendance check in at 2023-04-10 5am+07, UTC converted will be 2023-04-09 10pm
  - So the search does not take my attendance into account, I'm not sure whether it's matched with the module's purposes or not?

- The cast `::date` in the [sql query](https://github.com/OCA/hr-attendance/blob/15.0/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py#L57) outputs the date only, but the date is not at user's timezone. It is on-purpose and to match with the native report?


santostelmo (approved):

Highcooley (approved):
Functionally approved

victoralmau:
Please, cherry-pick https://github.com/OCA/hr-attendance/pull/136 to commit history

dzungtran89:
> Please, cherry-pick #136 to commit history

Hello @victoralmau , the PR has been updated

hr_attendance_report_theoretical_time/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Theoretical vs Attended Time Analysis",
+    "version": "16.0.1.0.0",
+    "category": "Human Resources",
+    "website": "https://github.com/OCA/hr-attendance",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": ["hr_attendance", "hr_holidays_public", "test_resource"],
```
victoralmau: `test_resource` is not a dependency that I like, but I understand that it is needed now.

victoralmau (approved):
Code review OK

hr_attendance_report_theoretical_time/__manifest__.py
pedrobaeza: This is not correct. You can't depend on a test module. Please explain the reasons, and if needed, then you should create an extra module `test_hr_attendance_report_theoretical_time` for that.

dzungtran89: Hello @pedrobaeza , the reason behind is `TestResourceCommon` has been moved to a dedicated module https://github.com/OCA/hr-attendance/pull/121/commits/df6b8e894bcabe5d081b8350167ed3600fa608a7#diff-d4983130c97220a392bdca5e411cb70875037e6e877ecc3f5f79f5e43707c596R9

dzungtran89: I will create a new test module to manage that change, thanks for your advise

pedrobaeza: Or you can just simply duplicate the code you need for your test in the `setUpClass`. Check the number of lines you need, and probably it will be the best option.

victoralmau (approved):

PR #128, opened 2023-05-16
[FIX] handle Validation error

lubusax:
If a ValidationError arises during the processing of an attendance, you must use str(e) in order to get the error message. Using only e gives later problems on other modules that can not handle with Objects of type ValidationError


OsoTranquilo (approved):

AaronHForgeFlow (approved):

AaronHForgeFlow:
ping @OCA/human-resources-maintainers

PR #129, opened 2023-05-26
[FIX] create and write function of res.company

yweng8111:
I've found an error in code.


yweng8111:
Sorry it is my own errors.

PR #137, opened 2023-07-24
[14.0][FIX] hr_attendance_sheet - Allows writing by batch

FrancoMaxime:
If you try to update a batch of hr.attendance.sheet, there is an error because operations are performed in a context of ensure_one(). it is therefore necessary to loop on the records

PR #138, opened 2023-07-24
[14.0] [ADD] hr_attendance_sheet_compensatory 

FrancoMaxime:
This module is based on hr_attendance_validation. It implements the same feature, but use the validation sheet of hr_attendance_sheet instead of creating a new model.
In this way, you can use both modules together instead of having two type of validation sheet containing the same data.

This allows to generate compensatory hours (allocated leaves) that can be used later as day off or to regulate credits leaves as this module is compatible with `hr_holidays_credit` module.

This PR needs the fix from #137


hr_attendance_sheet_compensatory/models/hr_attendance.py
```diff
@@ -0,0 +1,94 @@
...
+    is_overtime_due = fields.Boolean(
+        string="Is overtime due",
+        default=False,
+        help="Whether the overtime is due or not. "
+        "By default overtime is not due until a manager validates it.",
+    )
+
+    @api.model
+    def create(self, *args, **kwargs):
+        attendance = super().create(*args, **kwargs)
```
baimont: this feels weird

why isn't it this instead?

```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        attendances = super().create(vals_list)
```

hr_attendance_sheet_compensatory/models/hr_attendance.py
```diff
@@ -0,0 +1,94 @@
...
+                    "has already been reviewed and validated."
+                )
+                % (
+                    attendance.employee_id.name,
+                    attendance.check_in.date(),
+                )
+            )
+        return attendance
+
+    def unlink(self, *args, **kwargs):
```
baimont: same as above

hr_attendance_sheet_compensatory/models/hr_attendance.py
```diff
@@ -0,0 +1,94 @@
...
+                        "which would be moved to a validated day."
+                    )
+                    % (
+                        record.employee_id.name,
+                        record.check_in.date(),
+                    )
+                )
+        return res
+
+    def _is_validated_employee_week(self):
```
baimont, line 82: instead of using this in create and write, wouldn't it be better to define a constrains

hr_attendance_sheet_compensatory/views/hr_attendance_sheet.xml
```diff
@@ -0,0 +1,106 @@
...
+            <field name="overtime" position="replace" />
+            <field name="override_auto_lunch" position="after">
+                <field name="is_overtime" />
+                <field name="is_overtime_due" />
+            </field>
+            <field name="attendance_ids" position="attributes">
+                <attribute name="options" />
+            </field>
+            <notebook position="inside">
+                <page string="Leaves">
```
baimont: it would be nice to have name attributes on pages for inheritance purposes

hr_attendance_sheet_compensatory/views/hr_attendance_sheet.xml
```diff
@@ -0,0 +1,106 @@
...
+                                <field name="number_of_days" />
+                                <field
+                                    name="number_of_hours_display"
+                                    widget="float_time"
+                                />
+                            </tree>
+                        </field>
+                    </group>
+                </page>
+                <page string="Hours Information">
```
baimont: it would be nice to have name attributes on pages for inheritance purposes

douglascstd:
When this new feature become "Open" I'd like to validate (funcional review) it. Its a great increase to hr.attendance ;-) Good job @FrancoMaxime

FrancoMaxime:
@douglascstd Hi, sorry for late answer. I'm still working on hit, to fix small issues. I will send you a message when everything is fixed

hr_attendance_sheet_compensatory/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "HR Attendance Sheet Compensatory",
+    "version": "14.0.1.0.3",
+    "category": "Human Resources",
+    "summary": "Group attendances into attendance sheets.",
+    "website": "https://github.com/OCA/hr-attendance",
+    "author": "Odoo S.A., Odoo Community Association (OCA)",
```
petrus-v, line 11: Odoo S.A., really or missed copy past ?

hr_attendance_sheet_compensatory/models/hr_attendance.py
```diff
@@ -0,0 +1,98 @@
...
+        validated_week = (
+            self.env["hr.attendance.sheet"]
+            .with_user(SUPERUSER_ID)
+            .search_count(
```
petrus-v, line 86: this is not a best practice in the original code too, prefer sudo()

PR #139, opened 2023-08-07
[ADD] hr_attendance_overtime_from_work_calendar

hbrunn:
This module replaces Odoo core's overtime calculation with one that satisfies the expectation:

    overtime = (time recorded in hr.attendance records) - (expected working hours as per working calendar)

for every given day.

Note this is a draft and as of yet rather an invitation for discussion. The rough idea is:

- for every attendance, calculate the amount of working hours for the date with Odoo's resource calendar api (this way we get support for whatever module messes with work calendars for free), compute overtime as the difference between that and the worked hours from the attendances
- have a cronjob that creates overtime records for days without attendances as -(hours from the work calendar)

Currently this is super slow, but `_compute_missing_overtime` can be sped up considerably with some sql and a bit more convoluted grouping. It might also be worthwhile to have a start date for recalculations that admins can move forward periodically.

Further plan is to add `hr_attendance_overtime_from_work_calendar_demo` that depends on `hr_holidays` + `hr_holidays_public` and adds a bunch of (corner) test cases for this, and fix whatever comes out there.


albig:
Thank you for this work.

Just some notes after my first testing:

* the calculation of missing attendances works correctly (calculating the theoretical working time for a day)
* If the employee has a leave type holiday or sickness a duration of 0 is calculatet. That's fine.
* If the employee has a confirmed / validated leave of type "Extra Hours" the duration is 0, but no adjustment is calculated.
* Allocations of future "Extra Hours" leaves are not stored in the `hr_attendance_overtime` table anymore. Right?
* Allocation for sickness in the past (common usecase, employee comes back after sickness and fixes his attendance sheet), won't be fixed, because there is already an overtime entry for this date.

hbrunn:
we can have what this does simpler and less invasive with https://github.com/OCA/hr-attendance/pull/140 and https://github.com/OCA/hr-holidays/pull/87

PR #142, opened 2023-08-21
[ADD] hr_attendance_break

hbrunn:
This module lets employees take breaks on the attendance screen:

Screenshot_20230904_171541

This allows them to check in at the beginning of the working day, check out at the end and record breaks in between.

To be sure employees take enough breaks, there's also flagging for employees who didn't take enough breaks.


hbrunn:
@rafaelbn I'm curious what Spanish law says about mandatory work breaks? Plan here is to have the checkout screen scream if you worked longer than is allowed, and offer to adjust the break hours field accordingly.

For what my DE customer needs a minimum length for a break (15min) and a table

|Work hours|Min break|
|---------------|--------------|
| 6                | 30            |
| 9                | 45            |

would suffice, meaning if you work 6h or longer, you have to take half an hour break, if you work 9h or longer, you have to take at least three quarters. Would this be general enough to cover ES too?

albig:
In my setup, ordinary employees cannot access the attendance module with `hr_attendance_break` installed:



Maybe my rights are too strict? The employees should not get the kiosk widget and only have "Manual Attendance" rights.

hbrunn:
thanks, I keep getting fooled by the demo user which I use for unprivileged tests being actually quite privileged in HR

albig:
Thank you. Now it works for ordinary employees.

albig:
Two other issues:

- if the configured, minimum break length is 00:15, only breaks **_greater_** 15 minutes are taken into account and save to `break_hours`.
- if the employee has only 20 minutes break, but 30 minutes is the minimum break after 6 hours, the cron adds a break of 10 minutes. This break will not be saved in `break_hours`.

hbrunn:
1. my goal here is that only what we can legally count as break ends up in break_hours. How should I handle the minimum break otherwise then?
2. I see the problem

Should I then keep two fields, `break_hours_real` and `break_hours_legal`? Then you can report over both, employees see both and it's transparent where any divergence goes?

And we make also short breaks count towards the total when they have reason `hr_attendance_break.reason_imposed`?

albig:
1. It's fully correct to handle shorter breaks like this. I had the problem, that ab break of `00:15` was not accepted with a configured break of `00:15`. The problem are the seconds, we cannot (and don't want! to) see in the Odoo backend.
A break from `2023-09-18 12:02:18`	to `2023-09-18 12:17:17` is shown an `2023-09-18 12:02`to `2023-09-18 12:17`.
--> Would it be possible to truncate the seconds?
2. This would be a solution. Or just extend the "real" (e.g. 20 minutes) break with the "legal" break diff (e.g. 10 minutes)?


2023-09-18 12:02:18 | 2023-09-18 12:17:17
-- | --

albig:
If you click on break and try to stop the break again immediately, this will fail. It seems, that breaks < 1 minute cannot be interrupted.

hbrunn:
that's because your custom code changes the resolution of this field to 1 minute - not sure how to deal with that yet

albig:
Ahh, ok. Than this is no issue of this module. And maybe not that important.

PR #143, opened 2023-08-29
[15.0][FIX] hr_attendance_report_theoretical_time: Show the Reports menu to a basic user

victoralmau:
Show the Reports menu to a basic user (without Attendance > Officer group)

We need to leave the action empty for consistency because we are going to set different submenus, otherwise the Reports menu would not be displayed to a basic user.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT44849

### DIFF
--- a/hr_attendance_report_theoretical_time/i18n/es.po
+++ b/hr_attendance_report_theoretical_time/i18n/es.po
@@ -6,26 +6,31 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-11-14 13:18+0000\n"
-"PO-Revision-Date: 2021-01-25 15:44+0000\n"
+"POT-Creation-Date: 2023-08-31 06:15+0000\n"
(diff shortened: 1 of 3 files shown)


pedrobaeza:
How it was in v13?

victoralmau:
In v13 this fixed is not applied because it seems not to be necessary according to the access to the menus.

In v13 the objective is this:
v13-example

With these change, in v15 the operation is the same.

pedrobaeza:
But it's showing all the employees times or only the ones from your own?

victoralmau:
According to the defined rules (https://github.com/OCA/hr-attendance/blob/15.0/hr_attendance_report_theoretical_time/security/hr_attendance_report_theoretical_time_security.xml#L9), only your own rules can be displayed.

pedrobaeza:
Then I think the menu should be renamed to a more neutral one, as "All Employees" is not correct. Maybe just "Analysis"? Any other suggestion, @chienandalu ?

victoralmau:
Changed text to "Analysis".

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #148, opened 2023-10-09
[FIX][14.0] hr_attendance_rfid - Use a with savepoint instead of catching the error allowing a rollback

FrancoMaxime:
If an Error arises during the processing of an attendance, as it is not raised and just logged there is no rollback after. This is why I use a with self.env.cr.savepoint statement, to force a rollback in case of error trigger.

### DIFF
--- a/hr_attendance_rfid/i18n/fr.po
+++ b/hr_attendance_rfid/i18n/fr.po
@@ -4,17 +4,16 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 12.0\n"
+"Project-Id-Version: Odoo Server 14.0+e\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
This is anti-performant due to the creation of the savepoints, which scales very bad, but this code itself doesn't contain any `raise`, so it shouldn't be a problem. Shouldn't you act on the one generating the exception?

FrancoMaxime:
Hi @pedrobaeza , the problem here is inside the except who catch the error without any raise. If an error is triggered, the except will catch it, adds an error message in res. But all created data will be kept in database while there would have been a rollback if the error was raised correctly. This means that for odoo as the error is catch everything is ok, but in reality we created a  unstable record

pedrobaeza:
This method is not registering anything on the DB, so as said, you should the changes on the module where is doing such writing.

PR #156, opened 2023-11-16
uploaded  german translation

Raidan1200:
german translation


pedrobaeza:
Please use Weblate instead: https://odoo-community.org/page/translate

PR #159, opened 2024-01-16
[FIX] hr_attendance_report_theoretical_time: invalid recomputation of…

StephaneMangin:
… leaves using the wizard


pedrobaeza (changes requested):
Please indicate steps to reproduce the problem. What you are doing is very heavy computation to be added lightly.

StephaneMangin:
> Please indicate steps to reproduce the problem. What you are doing is very heavy computation to be added lightly.

Hi @pedrobaeza 

I draft it. It was not meant to be reviewable. Sorry for the noise.

PR #169, opened 2024-07-04
[16.0] [MIG] hr_birthday_welcome_message: Migration to 16.0

luisDIXMIT:
Standard migration.

- Added CONFIGURE.rst and USAGE.rst readme files.

@etobella


etobella (approved):
Still depends on Old JS, but the attendance app is done with the old Odoo JS Framework... :disappointed:

PR #171, opened 2024-07-16
raise AccessError if no employee is found for current user

Ricardoalso:
AccessError to be raised if no employee is found matching the domain, providing a clear error message and avoiding `ValueError: Expected singleton: hr.employee()`

### DIFF
--- a/hr_attendance_overtime/models/hr_employee.py
+++ b/hr_attendance_overtime/models/hr_employee.py
@@ -5,7 +5,8 @@ from datetime import datetime, time
 from dateutil.relativedelta import relativedelta
 from pytz import timezone, utc
 
-from odoo import SUPERUSER_ID, api, fields, models
+from odoo import SUPERUSER_ID, _, api, fields, models
+from odoo.exceptions import UserError
 from odoo.osv import expression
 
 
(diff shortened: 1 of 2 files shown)


hr_attendance_overtime/models/hr_employee.py
```diff
@@ -17,6 +18,9 @@ def todays_working_times(self, empl_domain):
         """Method used by my attendance/kiosk view in order
         to display employee planning and working times"""
         employee = self.search(empl_domain)
+        if not employee:
+            raise AccessError(_("Employee not found or not created for current user"))
```
petrus-v: I think it's not matter of access control raising `odoo.exceptions.UserError` is probably better here.

Also could you add a unit test to cover this case as suggested by codecov ?

victorvermot (approved):

petrus-v (approved):

hr_attendance_overtime/models/hr_employee.py
petrus-v: thanks taking care of this

petrus-v:
just realise as maintenaire of this module I could merge this PR before...

/ocabot merge patch

PR #193, opened 2024-11-27
[18.0][MIG] hr_attendance_reason

TomTietze:

PR #201, opened 2024-12-27
[18.0][MIG] hr_attendance_autoclose: Migration to 18.0

BhaveshHeliconia:
Depends on 

- https://github.com/OCA/hr-attendance/pull/193


AaronHForgeFlow:
This module maybe is deprecated or at least some refactor is needed because part of the feature is in Odoo standard now: https://github.com/odoo/odoo/blob/18.0/addons/hr_attendance/models/hr_attendance.py#L693

BhaveshHeliconia:
> This module maybe is deprecated or at least some refactor is needed because part of the feature is in Odoo standard now: https://github.com/odoo/odoo/blob/18.0/addons/hr_attendance/models/hr_attendance.py#L693

@AaronHForgeFlow Yes, you are right, we need to deprecate this module.

image

This is the default Odoo functionality.

PR #207, opened 2025-02-05
[15.0][FIX] hr_attendance_geolocation: correctly display seconds

pcastelovigo:
Currently seconds are shown with the minute symbol, making it difficult to operate with the values, e.g. copy and paste them into Google Maps.


pedrobaeza:
The same comment as for the 16.0 version. It's recommended that you do the patch in one version, and when approved, then spread across the rest of the versions, for avoiding having to patch on each PR if there's any comment (like this case).

PR #211, opened 2025-02-19
Update es.po

terralogia:
I have translated some missing messages


pedrobaeza:
This should be done in Weblate in https://translation.odoo-community.org/

Please check https://odoo-community.org/resources/translate and use "trato de usted", no "trato de tú".

PR #215, opened 2025-03-14
[17.0][ADD] hr_attendance_geolocation_required

Arubaru86:
This module ensures that geolocation is required for employee check-ins and check-outs
in the Odoo attendance system. Employees must have GPS enabled and allow location 
permissions in their browser before they can clock in or out.

### Features:
- Prevents attendance if geolocation permissions are denied.
- Displays a warning message when GPS is not enabled.
- Ensures that all attendances include latitude and longitude.
- Works in **Kiosk Mode** and **Manual Mode**.
- Extends `public_kiosk_app.js` to enforce geolocation checks.

### Impact:
- Guarantees location-based attendance tracking.
- Improves data accuracy for remote workers.
- Ensures compliance with location-based attendance policies.


torojur25 (approved):

hr_attendance_geolocation_required/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Grupo Isonor
+# Copyright (C) All Rights Reserved
+
+{
+    "name": "HR Attendance Geolocation Required",
+    "summary": "Hace obligatoria la geolocalización para fichar en Odoo 17",
+    "author": "Odoo Community Association (OCA),Grupo Isonor, Álvaro Alonso",
```
Tardo:
```suggestion
    "author": "Grupo Isonor, Odoo Community Association (OCA)",
```

hr_attendance_geolocation_required/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "website": "https://github.com/OCA/hr-attendance",
+    "license": "LGPL-3",
+    "category": "hr_attendance",
+    "version": "17.0.1.0.0",
+    "depends": ["hr_attendance"],
+    "data": [],
+    "assets": {
+        "web.assets_backend": [
+            "hr_attendance_geolocation_required/static/src/public_kiosk/public_kiosk_app.js",
```
Tardo, line 17:
```suggestion
            "hr_attendance_geolocation_required/static/src/public_kiosk/public_kiosk_app.esm.js",
```

Tardo (changes requested):
Please, add tests... and the 'readme' folder with "CONTRIBUTORS, DESCRIPTION, ..." files.

victoralmau (commented):
I think another one could be to add this directly in `hr_attendance_geolocation` through a boolean field at `res.config.settings` level and activate it who needs it. What do you think?

pedrobaeza:
AFAIK, `hr_attendance_geolocation` is no longer needed in 17, as it's included in core. Can you confirm?

victoralmau:
> AFAIK, `hr_attendance_geolocation` is no longer needed in 17, as it's included in core. Can you confirm?

You're right, also I had it noted internally (I didn't remember it in the quick comment above) but it looks like OpenUpgrade is missing, I'll add it soon to avoid confusion.

https://github.com/OCA/OpenUpgrade/pull/4693

PR #218, opened 2025-03-31
[16.0][ADD] hr_attendance_gantt_view

MarinaAForgeFlow:


hr_attendance_gantt_view/views/hr_attendance_views.xml
```diff
@@ -0,0 +1,47 @@
...
+                            />
+                            <t t-esc="hours" /> hours and <t t-esc="minutes" /> minutes
+                        </div>
+                    </div>
+                </templates>
+            </gantt>
+        </field>
+    </record>
+    <record id="hr_attendance.hr_attendance_action" model="ir.actions.act_window">
+        <field name="view_mode">tree,kanban,form,gantt</field>
```
MarinaAForgeFlow, line 45: I am not sure if here is any way to inherit an ir.actions.act_window and add a view mode to the "view_mode" field without overwriting the originals. Any suggestion?

hr_attendance_gantt_view/views/hr_attendance_views.xml
```diff
@@ -0,0 +1,47 @@
...
+                            />
+                        </div>
+                        <div>
+                            <strong>Worked Hours: </strong>
+                            <t t-set="hours" t-value="Math.floor(worked_hours)" />
+                            <t
+                                t-set="minutes"
+                                t-value="Math.round((worked_hours - hours) * 60)"
+                            />
+                            <t t-esc="hours" /> hours and <t t-esc="minutes" /> minutes
```
MarinaAForgeFlow, line 37: Could not find the correct way to format the worked_hours float as a time_float inside the template. Would really appreciate suggestions in which is the better and cleanest way to do it.

hr_attendance_gantt_view/views/hr_attendance_views.xml
JordiMForgeFlow: @MarinaAForgeFlow AFAIK this is fine :)

PR #219, opened 2025-04-08
[17.0][MIG] hr_attendance_overtime

petrus-v:
I'm not user odoo overtime because this is base module for hr_attendance_validation which works on a week to create hr leave allocation while odoo is working on a day.

- [x] migrate models and views
- [ ] migrate check-in/check-out frontend component


hr_attendance_overtime/models/resource.py
```diff
@@ -0,0 +1,116 @@
...
+    #     res = super()._copy_attendance_vals()
+    #     res.update(
+    #         {
+    #             "hour_check_in_from": self.hour_check_in_from,
+    #             "hour_check_in_to": self.hour_check_in_to,
+    #             "hour_check_out_from": self.hour_check_out_from,
+    #             "hour_check_out_to": self.hour_check_out_to,
+    #         }
+    #     )
+    #     return res
```
petrus-v, line 101: to be remove if not usefull !

PR #226, opened 2025-07-01
[18.0][MIG] hr_attendance_rfid: Migration to version 18.0

david-banon-tecnativa:
Minor updates only
@tecnativa @victoralmau @pedrobaeza 

TT54168


hr_attendance_rfid/tests/test_hr_attendance_rfid_process.py
```diff
@@ -66,7 +65,8 @@ def test_invalid_code(self):
         """Invalid employee"""
         self.employee_model = self.employee_model.with_user(self.env.user)
         invalid_code = "029238d"
-        res = self.employee_model.register_attendance(invalid_code)
+        with mute_logger("odoo.addons.hr_attendance_rfid.models.hr_employee"):
```
pedrobaeza: Which warning is appearing here?

david-banon-tecnativa: https://github.com/OCA/hr-attendance/blob/f45400349fd4972b36f14842906d58bf2f1920e7/hr_attendance_rfid/models/hr_employee.py#L53

hr_attendance_rfid/tests/test_hr_attendance_rfid_process.py
```diff
@@ -0,0 +1,86 @@
...
+                    datetime.today() + timedelta(hours=8)
+                ),
+            }
+        )
+        employee.update({"attendance_state": "checked_in"})
+        res = self.employee_model.register_attendance(self.rfid_card_code)
+        self.assertNotEqual(res["error_message"], "")
+
+    @users("hr_attendance_rfid-user")
+    def test_invalid_code(self):
```
victoralmau:
```suggestion
    @users("hr_attendance_rfid-user")
    @mute_logger("odoo.addons.hr_attendance_rfid.models.hr_employee")
    def test_invalid_code(self):
```

hr_attendance_rfid/tests/test_hr_attendance_rfid_process.py
```diff
@@ -0,0 +1,86 @@
...
+        """Invalid employee"""
+        self.employee_model = self.employee_model.with_user(self.env.user)
+        invalid_code = "029238d"
+        with mute_logger("odoo.addons.hr_attendance_rfid.models.hr_employee"):
```
victoralmau:
```suggestion
```

victoralmau (changes requested):

hr_attendance_rfid/tests/test_hr_attendance_rfid_process.py
david-banon-tecnativa: I think its better to have the mute self contained, as we only want to ignore that warning because we are testing the code that causes the warning.

victoralmau: But...in v17 there is something similar in another test https://github.com/OCA/hr-attendance/blob/fec5cef610b124f76bc438d388cfda4c919e5198/hr_attendance_rfid/tests/test_hr_attendance_rfid_process.py#L46

victoralmau (approved):
Code review OK

JordiBForgeFlow (approved):

PR #228, opened 2025-07-28
[18.0][MIG] hr_attendance_autoclose > hr_attendance_flexible_autoclose: Migration to 18.0

Kimkhoi3010:
Migrate `hr_attendance_autoclose` to 18.0

### Change
- Handle employees with flexible_hours = True.
- Use a single unified cron to avoid overlapping or conflicting logic.
- Rename to `hr_attendance_flexible_autoclose`


victorvermot:
@Kimkhoi3010 Hello ! In previous versions of the module, there was an attendance reason that the user could set as a setting. It was then added to auto closed attendances. Any reason to have left that logic out of this pr ?

cyrilmanuel:
@Kimkhoi3010  PIng :)

cyrilmanuel (approved):

hr_attendance_flexible_autoclose/models/hr_attendance.py
```diff
@@ -0,0 +1,87 @@
...
+from dateutil.relativedelta import relativedelta
+
+from odoo import fields, models
+
+
+class HrAttendance(models.Model):
+    _inherit = "hr.attendance"
+
+    def _cron_auto_check_out(self):
+        super()._cron_auto_check_out()
```
petrus-v, line 16: as far I understand this will call odoo source code so the code bellow becomes useless as the [Odoo code](https://github.com/odoo/odoo/blob/18.0/addons/hr_attendance/models/hr_attendance.py#L707) will already close it, isn't it ?


This module is deprecated OpenUpgrade consider it's not need anymore https://github.com/OCA/OpenUpgrade/pull/5127

petrus-v (commented):
I suppose this module should not be migrated or we should  rename the module to avoid confusion if we needs to improve Odoo core module

hr_attendance_flexible_autoclose/models/hr_attendance.py
jans23: Odoo's core automatic check out doesn't work for flexible working hours but hr_attendance_autoclose does. Thus I think hr_attendance_autoclose is still useful and required for such cases.

KDRdev: It's still useful, as Odoo's code doesn't account for flexible hours, but this module does.

KDRdev (approved):

cyrilmanuel (approved):

cyrilmanuel:
hi @pedrobaeza could you merge this pr please :)

pedrobaeza:
This comment should be attended AFAIK (about the renaming): https://github.com/OCA/hr-attendance/pull/228#pullrequestreview-3655048991

jans23:
@Kimkhoi3010 do you have interest in moving this forward and renaming the module, as suggested?

Kimkhoi3010:
> @Kimkhoi3010 do you have interest in moving this forward and renaming the module, as suggested?

Hi @jans23, @cyrilmanuel, @victorvermot, @pedrobaeza, thanks for your time. I have updated the PR

jans23:
I'm not sure but should the PR be renamed too (also ADD)?

pedrobaeza:
You should rename whole commit history with:

```
git filter-branch --tree-filter 'if [ -d hr_attendance_autoclose ]; then mv hr_attendance_autoclose hr_attendance_flexible_autoclose; fi' HEAD
git rebase origin/18.0
```

or if the git version is newer:

```
git filter-repo --path-rename hr_attendance_autoclose/:hr_attendance_flexible_autoclose/ --refs HEAD --force
git rebase origin/18.0
```

Kimkhoi3010:
> You should rename whole commit history with:
> 
> ```
> git filter-branch --tree-filter 'if [ -d hr_attendance_autoclose ]; then mv hr_attendance_autoclose hr_attendance_flexible_autoclose; fi' HEAD
> git rebase origin/18.0
> ```
> 
> or if the git version is newer:
> 
> ```
> git filter-repo --path-rename hr_attendance_autoclose/:hr_attendance_flexible_autoclose/ --refs HEAD --force
> git rebase origin/18.0
> ```

Thanks @pedrobaeza, I have update the PR.

pedrobaeza:
ping @victoralmau

hr_attendance_flexible_autoclose/models/hr_attendance.py
victoralmau: I have proposed that Odoo take flexible working hours into account (the existing code in the method is correct, in my opinion; we would just need to remove the corresponding domain): https://github.com/odoo/odoo/pull/282685

jans23:
@victoralmau could you review this, please?

victoralmau:
I agree with this module, but it frustrates me a little that I need it **ONLY** for flexible calendars; I created a pull request for Odoo (https://github.com/odoo/odoo/pull/282685) to add that compatibility (in my opinion, nothing extra is really needed for compatibility), but it’s still pending; the best thing would be for Odoo to merge that PR so that this module is no longer necessary.

PR #229, opened 2025-08-04
[17.0][ADD] hr_attendance_resource_calendar

edescalona:
@BinhexTeam

This module allows you to block attendance based on the employee's work schedule.

Allows you to define a time value which allows the employee to check in before the actual time.


hr_attendance_resource_calendar/models/hr_employee.py
```diff
@@ -0,0 +1,56 @@
...
+        float_time = now_tz.hour + now_tz.minute / 60 + now_tz.second / 3600
+        is_work_day = self.env["resource.calendar.attendance"].search(
+            [
+                ("calendar_id", "=", self.resource_calendar_id.id),
+                ("dayofweek", "=", day_week),
+            ]
+        )
+        if is_work_day:
+            min_hour_from = min(is_work_day.mapped("hour_from"))
+            max_hour_to = max(is_work_day.mapped("hour_to"))
```
christian-ramos-tecnativa: At first this seems to be more efficient but the problem appear in situations like this one:
image
With this configuration and how is developed right now it will allow the user to log in between 12:00 and 13:00 wich is incorrect

christian-ramos-tecnativa (changes requested):

christian-ramos-tecnativa:
Also I think it would be nice to have a variable value configurable, so the employees are able to log in 10 or 5 minutes before the actual time (or any value set by the user). It could be added in the calendar as a new field.

edescalona:
Hi @Christian-RB ready for review

christian-ramos-tecnativa:
@edescalona Why is the restriction only applied to the first check-in of the day?
In my opinion, it should apply to every entry. Otherwise, what will likely happen is that all employees will finish lunch and then queue up to check in at the same time, meaning the last person in line will end up checking in too late, even if they returned on time. Let me know is I explain my self properly. Thanks!

christian-ramos-tecnativa:
Also, to ensure the tests pass, you should add a configurable setting on the calendar or `res.config.settings` so the original workflow works as usual.

edescalona:
Hi @Christian-RB  thanks for your comments, let me know when you can.

PR #231, opened 2025-08-18
[18.0][MIG] hr_attendance_autoclose: Migration to 18.0

victorvermot:


pedrobaeza:
This module is no longer needed: https://github.com/OCA/OpenUpgrade/pull/5127

PR #233, opened 2025-10-20
[ADD] hr_attendance_block_by_geolocation: New module hr_attendance_block_by_geolocation

miquelpascual:
This module allows you to define zones for clocking in. 
If an employee is not within one of these zones, they will not be able to clock in. 
A boolean field is also added to allow them to clock in from any location.

cc https://github.com/APSL 22326
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @BernatObrador @ppyczko please review


miquelalzanillas:
Hi @mpascuall 

IMO a better name for this module would be, 'hr_attendance_block_by_geolocation'.

Some tests for his basic flows are missing too and IMO are important in this module.

peluko00 (approved):
LGTM! Code review and tested in runboat

hitrosol:
It works in my testing DB. 
I wonder how about if particular employee can only check-in/out for not all authorized locations ? or depends on the office location ? I think this will be good enhancement in the future...

miquelpascual:
> It works in my testing DB. I wonder how about if particular employee can only check-in/out for not all authorized locations ? or depends on the office location ? I think this will be good enhancement in the future...

Hi, thanks for your review! The module is designed to create a zone —usually the office— where employees can record their attendance. Your idea of limiting the zone by employee is a good one; it could be a nice improvement for the future!

lbarry-apsl (approved):
Code and runboat reviewed. LGTM

PR #234, opened 2025-10-21
[17.0][MIG] hr_attendance_overtime

petrus-v:
replace #219 

I'm not using odoo overtime because this is base module for hr_attendance_validation which works on a week to create hr leave allocation while odoo is working on a day.

- [x] migrate models and views
- [x] migrate check-in/check-out frontend component

PR #235, opened 2025-11-25
[ADD] hr_overtime_report: New module hr_overtime_report

u0f:
This modulle allows you to see the full overtime report including attendances,  absences, and manual adjustments. 
Restores the complete overtime from v16 under HR \> Reports \> Extra Hours, showing all overtime sources with list, pivot, graph, and calendar views.

https://www.loom.com/share/49580ed6428b4e1487d53b61e44d67ed

MT-12632 @moduon @yajo


hr_overtime_report/readme/CONTEXT.md
```diff
@@ -0,0 +1,2 @@
+This module restores the complete overtime from v16 under HR \> Reports \> Extra Hours, 
+showing all overtime sources with list, pivot, graph, and calendar views.
```
yajo: issue: don't refer to the past.

In general terms, users reading a README don't really care about the past. They're reading a README for Odoo 18, and they care about what this module adds to their Odoo.

Not referring to past version also helps maintaining the README. 10 years after now, when you are training a newbie to migrate it for Odoo 28, nobody will care about what Odoo 16 did.

So, this is a better context:


```suggestion
When an HR manager tries to get extra hours information of their employees,
Odoo will only display information related to attendances.

However, there can be extra hour records related to time off allocations or
requests. That information is also relevant, but is hidden by default.

HR managers that extra hours information of their employees will appreciate
the new menu that this module provides, where they will find that information.
```

hr_overtime_report/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    "name": "HR Overtime Report",
+    "summary": "Overtime report in Employees",
+    "version": "18.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Human Resources",
+    "website": "https://github.com/OCA/hr-attendance",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": [],
```
yajo: issue: add maintainers


```suggestion
    "maintainers": ["rafaelbn", "u0f"],
```

hr_overtime_report/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "hr",
```
yajo: issue: wrong dependencies.

IIUC, the `hr.attendance.overtime` is defined in `hr_attendance` instead:


```suggestion
        "hr_attendance",
```

yajo (changes requested):
Very good job!

loida-vm (changes requested):
The Adjustment field in the list view is activated when a new daily report is manually created for an employee, indicating how many overtime hours they have worked.
However, this does not make much sense, because such reports should not be created manually.
These reports are generated automatically from the recorded attendance entries.

Therefore, the button "New" in the list view should not appear. This has already been talked, and the necessary changes will be made.
Thanks @u0f

u0f:
Thanks four your review @loida-vm!

- I removed the filters "From Attendance" and "Ajustments/Abscences" 
- I removed the field "Adjustment" from the list view
- Now is not posible create records
- Update translation and README

image

image

loida-vm (approved):
Thank you very much @u0f 
From my side, as a functional review, it's approved.
Regarding the modification of the overtime report, this is because if the attendance data is not correct, it must be adjusted directly in the attendance module, as has always been done. Therefore, this report remains exactly that: a report, and doesn't directly affect the attendance data.

LGTM! 😃

yajo (approved):

rafaelbn (changes requested):
Hello! Thank you for this work!

After review:

- [ ] Filter active employees by default
- [ ] Don't filter by date by default
- [ ] Delete "Extra hours (Real)". Is false sentence. That calculations comes from attendances without the extra hours compensations, so two reasons (1) More accurate is "extra hours (incomplete)" and this has no sense (2) less is more, clear is better than unclear 😄 
- [ ] Translate to Spanish and German

UX and psychology is important in software development , that's why it is difficult 🔝 ❤️  

Thank you!

imagen

u0f:
Thanks for your correction, @rafaelbn. I made the changes:
- Default filter is now "Active Employees"
- I deleted "Extra hours (Real)" from the report
- Translations for Spanish and German

image

yajo (changes requested):
> Delete "Extra hours (Real)". Is false sentence. That calculations comes from attendances without the extra hours compensations, so two reasons (1) More accurate is "extra hours (incomplete)" and this has no sense (2) less is more, clear is better than unclear 😄

The real difference is: is tolerance applied?

Here you configure tolerance:

image

If tolerance is 15 minutes, a record of 10 minutes will be: 0:00 extra hours (0:10 real extra hours).

I agree on hiding the field by default, but removing it can be problematic if you use tolerance IIUC.

loida-vm:
Please, let's remove all the views that don’t add any value. Let's leave it like this:
imagen
I explain it in the next video: https://www.loom.com/share/18b753f367ae4e88a8ae8a9941778802
Thank you!

u0f:
Requested change done, @loida-vm!

image

rafaelbn (approved):
Reviewed, tested and confirmed 👍🏼 

This is the best mvp simple and easy

loida-vm:
Thanks @u0f 
Everything looks perfect!

yajo (approved):
/ocabot merge nobump

u0f:
Hey @OCA/human-resources-maintainers, can someone merge this? Thanks!

rrebollo:
@OCA/human-resources-maintainers can someone merge this? Thanks in advance.

rrebollo:
@u0f what about the overtime hours from #234?

EmilioPascual (approved):

PR #240, opened 2025-12-22
[15.0][ADD] hr_attendance_rest_time_included: New module

eduezerouali-tecnativa:
cc @Tecanativa TT59859
ping @pedrobaeza @juancarlosonate-tecnativa @david-banon-tecnativa 

This module allows tracking employee rest times (breaks) that are included in the employee’s working time.


pedrobaeza (commented):
@eduezerouali-tecnativa you have to add at attendance reason level, when being "Sign out" one, a check for saying if it's a rest time included in the attendance.

pedrobaeza:
I see you have added it as a configuration option inside settings, but it should be at attendance reason level, as it can be more than one reason.

hr_attendance_rest_time_included/models/hr_attendance.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2025 Tecnativa - Eduardo Ezerouali
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import api, fields, models
+
+
+class HrAttendance(models.Model):
+    _inherit = "hr.attendance"
+
+    hr_attendance_rest_time = fields.One2many(
```
pedrobaeza: This is inside `hr.attendance`, so the prefix `hr_attendance` is useless, redundant and takes a lot of space, and on the other side, use `_ids` suffix convention for o2m:

```suggestion
    rest_time_ids = fields.One2many(
```

hr_attendance_rest_time_included/models/hr_attendance_rest_time.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2025 Tecnativa - Eduardo Ezerouali
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import fields, models
+
+
+class HrAttendanceRestTime(models.Model):
+    _name = "hr.attendance.rest_time"
+    _inherit = "hr.attendance"
```
pedrobaeza: For indicating the intention of inheritance with new model, better to put before `_inherit` and then `_name`:

```suggestion
    _inherit = "hr.attendance"
    _name = "hr.attendance.rest_time"
```

hr_attendance_rest_time_included/models/hr_employee.py
```diff
@@ -0,0 +1,38 @@
...
+from odoo import fields, models
+
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    def _attendance_action_change(self):
+        self.ensure_one()
+        RestTime = self.env["hr.attendance.rest_time"]
+
```
pedrobaeza: Don't add empty lines inside a method (it applies to all the code).

hr_attendance_rest_time_included/models/res_company.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2025 Tecnativa - Eduardo Ezerouali
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    hr_attendance_rest_time = fields.Boolean(string="Allow rest time included")
```
pedrobaeza: To be put at attendance reason level, not here.

hr_attendance_rest_time_included/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Eduardo Ezerouali
```
pedrobaeza, line 4: You should add me, being the designer of the solution and the main reviewer.

hr_attendance_rest_time_included/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module extends `hr_attendance` to manage employee rest times included in working time, supporting multiple breaks per attendance.
+
```
pedrobaeza: Remove the extra empty line.

hr_attendance_rest_time_included/report/hr_attendance_report_views.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record id="hr_attendance_report_view_pivot" model="ir.ui.view">
+        <field name="name">hr.attendance.report.view.pivot</field>
+        <field name="model">hr.attendance.report</field>
+        <field name="inherit_id" ref="hr_attendance.hr_attendance_report_view_pivot" />
+        <field name="arch" type="xml">
+            <xpath expr="//pivot" position="inside">
+                <field name="rest_hours" type="measure" widget="float_time" />
```
pedrobaeza: Is this really needed for the measure to be shown in the pivot?

hr_attendance_rest_time_included/models/hr_employee.py
```diff
@@ -0,0 +1,38 @@
...
+            if last_attendance and last_attendance.check_out:
+                last_attendance.write({"check_out": False})
+            return last_attendance
+        res = super()._attendance_action_change()
+        reason_id = self.env.context.get("attendance_reason_id")
+        if (
+            self.company_id.hr_attendance_rest_time_id
+            and self.company_id.hr_attendance_rest_time_id.id == reason_id
+        ):
+            RestTime.create(
```
pedrobaeza: I think this can be problematic with the current security permissions when a base employee with attendance permissions. I think you should add `hr_attendance.group_hr_attendance` to the permissions, or do this with `sudo`.

pedrobaeza (changes requested):
Apart from the comments inline, include in the ROADMAP that this module is not totally compatible with `hr_attendance_geolocation`, as I have seen that the sign out coordinates are stored in the main attendance record while signing out the rest time. On upper versions, where the geolocation is included, we will need to check how to make it compatible.

eduezerouali-tecnativa:
All changes apply !

hr_attendance_rest_time_included/models/hr_attendance_rest_time.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import fields, models
+
+
+class HrAttendanceRestTime(models.Model):
+    _inherit = "hr.attendance"
+    _name = "hr.attendance.rest_time"
+    _description = "Hr attendance break"
+
+    hr_attendance_id = fields.Many2one(
```
pedrobaeza: This is usually put this way, as it doesn't add too much value to put `hr_` as prefix, but this is not blocking:

```suggestion
    attendance_id = fields.Many2one(
```

hr_attendance_rest_time_included/views/hr_attendance_reason_views.xml
```diff
@@ -0,0 +1,29 @@
...
+    <record id="hr_attendance_reason_view_form" model="ir.ui.view">
+        <field name="name">hr.attendance.reason.form</field>
+        <field name="model">hr.attendance.reason</field>
+        <field
+            name="inherit_id"
+            ref="hr_attendance_reason.hr_attendance_reason_view_form"
+        />
+        <field name="arch" type="xml">
+            <field name="company_id" position="after">
+                <field name="include_rest_time" />
```
pedrobaeza: This should be shown only for sign out reasons.

hr_attendance_rest_time_included/models/hr_attendance_reason.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2025 Tecnativa - Eduardo Ezerouali
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import fields, models
+
+
+class HrAttendanceReason(models.Model):
+    _inherit = "hr.attendance.reason"
+
+    include_rest_time = fields.Boolean(default=False)
```
pedrobaeza: And the default is False already, so no need to put it.


```suggestion
    rest_time_included = fields.Boolean()
```

pedrobaeza (commented):
Apart from the inline comments, we are almost there. Some other changes:

- It seems the original chatter is being rewritten when checking out/in the rest time:

  Image

- There's no navigation nor visualization of the rest time attendances. Put a button next to the rest time column to go to them, and also a smart-button on the main attendance record, both only visible when there are rest time attendances.

eduezerouali-tecnativa:
Small comments already fix (not push yet). About the chatter, this is only available if you install `hr_attendance_modification_tracking` in upper odoo version is directly into core. Should we add a glue module to fix that?

pedrobaeza:
> About the chatter, this is only available if you install `hr_attendance_modification_tracking` in upper odoo version is directly into core. Should we add a glue module to fix that?

No need of the glue module then, but also a mention in the ROADMAP.

eduezerouali-tecnativa:
add changes, plus the mention to ROADMAP

pedrobaeza:
Last fix:

image

eduezerouali-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
Can you put another icon like the clock for the button to go the the breaks?

eduezerouali-tecnativa:
done

eduezerouali-tecnativa:
added Spanish translations, not 100% if it goes in an other commit or i should just squash them

pedrobaeza:
It should go in the same commit as the other, and anyway, if in a separate commit, it shouldn't be `[ADD]`, as that's just for adding modules, but squash it as said.

eduezerouali-tecnativa:
> It should go in the same commit as the other, and anyway, if in a separate commit, it shouldn't be `[ADD]`, as that's just for adding modules, but squash it as said.

okay, i got that from an other module commit as reference to add translation without using weblate. Anyways i just add it to the commit. Thank you!

hr_attendance_rest_time_included/views/hr_attendance_views.xml
```diff
@@ -0,0 +1,52 @@
...
+                    >
+                    <field name="rest_time_count" widget="statinfo" string="Breaks" />
+                </button>
+            </div>
+        </xpath>
+            <sheet position="inside">
+                <group name="breaks">
+                    <group>
+                        <field
+                            name="rest_time_ids"
```
pedrobaeza: Why adding this if it's invisible?

victoralmau (changes requested):
Code review OK

The comment https://github.com/OCA/hr-attendance/pull/240#discussion_r2661319820 is pending (if the field is invisible, it is not necessary to add it).

hr_attendance_rest_time_included/views/hr_attendance_views.xml
eduezerouali-tecnativa: actually it is not necessary

eduezerouali-tecnativa:
changes done that field was not necessary, did the change on the mig too

victoralmau (approved):

david-banon-tecnativa (approved):
Functional review OK, code LGTM

PR #222, opened 2025-04-30
[17.0][ADD] hr_attendance_report

Arubaru86:
This module allows users to generate employee attendance reports in PDF and Excel formats within the Odoo Attendance system. It provides a wizard for filtering and exporting attendance data with customizable date ranges, employees, and departments.

### Features:
- Generates PDF and Excel attendance reports.
- Filters attendance data by employee, department, month, and year.
- Displays detailed information, including check-in, check-out, and worked hours.
- Supports multiple employee and department selections.
- Integrates into the HR Attendance menu and enhances reporting capabilities.

### Impact:
- Simplifies the process of generating attendance reports for HR teams.
- Improves the efficiency of reporting for payroll or compliance purposes.
- Provides a more streamlined and accurate tracking of employee attendance data.
- Enhances visibility and reporting capabilities across departments.

PR #242, opened 2026-01-13
[17.0][ADD] hr_attendance_report

Neitherkx:
Supersede #222

This PR updates the Employee Attendance Reporting module (hr_attendance_report).

Original PR summary:
The module allows generating employee attendance reports in PDF and Excel formats within Odoo Attendance. It provides a wizard to filter by employee, department, month, and year, displaying detailed information such as check-in, check-out, and worked hours. It streamlines reporting for HR and payroll purposes.

Changes in this PR:
- Adds company and employee identification number to the reports.
- Improves the PDF view for clearer presentation.
- Adds wizard access permissions: administrators can select multiple employees, while regular users can only select themselves.


rrebollo:
I think you should squash the commits.

rrebollo:
Why there is changes not related to the hr_attendance_report included?

image

hr_attendance_report/report/employee_attendance_report_wizard_view.xml
rrebollo, line 1: Consider replace depreacted t-esc by t-out

hr_attendance_report/report/hr_report.py
```diff
@@ -0,0 +1,141 @@
...
+                "job_title": emp.sudo().job_id.name if emp.job_id else _("N/A"),
+                "attendances": attendance_data,
+                "total_hours": round(total_hours, 2),
+                "total_days": len(dates_worked),
+                "avg_hours_per_day": round(total_hours / len(dates_worked), 2)
+                if dates_worked
+                else 0,
+            }
+
+            employee_info_list.append(employee_info)
```
rrebollo, line 149: Attendance records could become a large dataset, so performance starts to matter. Instead of running a search per employee, why not perform a single search for all employees, then sort by employee_id, check_in? Alternatively, you could use grouped, read_group, or some of the newer ORM utilities, and then do the Python processing afterward.

hr_attendance_report/tests/test_attendance_pdf_report.py
```diff
@@ -0,0 +1,720 @@
...
+import datetime
+import logging
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+_logger = logging.getLogger(__name__)
+
+
+class TestAttendancePdfReport(TransactionCase):
```
rrebollo, line 13: Consider to use `BaseCommon` instead of `TransactionCase`

hr_attendance_report/tests/test_attendance_report_wizard.py
```diff
@@ -0,0 +1,735 @@
+# Copyright 2025 Álvaro Alonso Bada - Grupo Isonor
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import datetime
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestAttendanceReportWizard(TransactionCase):
```
rrebollo, line 10: Idem

hr_attendance_report/wizard/employee_attendance_report_wizard.py
```diff
@@ -0,0 +1,320 @@
...
+        default=False, string="Select All Departments"
+    )
+    select_month = fields.Selection(
+        MONTH_SELECTION,
+        string="Month",
+        required=True,
+        default=str(datetime.date.today().month),
+    )
+    select_year = fields.Char(
+        string="Year", required=True, default=str(datetime.date.today().year), size=4
```
rrebollo:
```suggestion
        string="Year", required=True, default=lambda self: fields.Date.today().year, size=4
```

Without `lambda`, this executes when the server starts, if I'm not mistaken. Also, you don't need `datetime` for this. In fact, I think almost everything done in this addon with `datetime` could be handled using the helper methods from the `fields.Date` and `fields.Datetime` classes.

hr_attendance_report/wizard/employee_attendance_report_wizard.py
```diff
@@ -0,0 +1,320 @@
...
+
+    @api.model
+    def _get_view(self, view_id=None, view_type="form", **options):
+        arch, view = super()._get_view(view_id, view_type, **options)
+        # Make employee field readonly for non-managers
+        if not self.env.user.has_group("hr_attendance.group_hr_attendance_manager"):
+            if view_type == "form":
+                for node in arch.xpath("//field[@name='hr_employee_ids']"):
+                    node.set("readonly", "1")
+        return arch, view
```
rrebollo: I think you could achieve the same thing with a less hacky approach in the view using the `groups` attribute:

```xml
<field name="employee_ids" readonly="1" groups="!hr_attendance.group_hr_attendance_manager" />
<field name="employee_ids" groups="hr_attendance.group_hr_attendance_manager" />
```

hr_attendance_report/wizard/employee_attendance_report_wizard.py
```diff
@@ -0,0 +1,320 @@
...
+        self, workbook, employee, start_date, end_date
+    ):
+        attendances = self.env["hr.attendance"].search(
+            [
+                ("employee_id", "=", employee.id),
+                ("check_in", ">=", start_date),
+                ("check_in", "<=", end_date),
+            ],
+            order="check_in",
+        )
```
rrebollo: This logic appears to be repeated, it could be refactored. Also, the search by employee issue I pointed out earlier.

rrebollo: I guess is one sheet by employee approach

rrebollo (changes requested):
Thanks for your contribution. Please remove files that are not related to the declared goal of the PR. Also, please consider the rest of my suggestions—most are not mandatory, but worth taking into account.

Neitherkx:
Hi @rrebollo, thanks for your feedback. I’ve been applying the changes you suggested, along with some additional improvements. Could you take a look when you have a moment?

rrebollo (approved):
Code Review. LGTM!

I'm curious. Why the move to xlsxwriter?

Neitherkx:
> Code Review. LGTM!
> 
> I'm curious. Why the move to xlsxwriter?

Hi! `xlwt` is deprecated and only supports the old `.xls` format. I switched to `xlsxwriter` since it generates modern `.xlsx` files and is already a native Odoo dependency, so no extra installation is needed.

Neitherkx:
Hello @pedrobaeza @yajo,

Tests are green and the PR seems ready from my side.  
Could you please take a look when possible?

Also, there is another related PR for the migration to 18.0 still pending review.
Thanks in advance!

PR #224, opened 2025-06-04
[17.0][WIP] hr_attendance_report_theoretical_time: Create empty attendances to improve report

victoralmau:
Create empty attendances to improve report

@Tecnativa

PR #247, opened 2026-01-30
[18.0][IMP] hr_attendance_report_theoretical_time: Create empty attendances

etobella:
Adapt to 18 and removed some unnecessary code that was the actual culript


etobella:
Tested in a Database with 1.000 employees and it works like a charm

victoralmau (approved):
Thank you for continuing what was started in https://github.com/OCA/hr-attendance/pull/224

Code review OK

In my opinion, this approach is appropriate, consistent, and “simple.”

victoralmau:
Ping @pedrobaeza can you review it?

victoralmau:
Ping @juancarlosonate-tecnativa and @Andrii9090-tecnativa can you review it?

Andrii9090-tecnativa (approved):

victoralmau:
Ping @pedrobaeza Can this be merged?

pedrobaeza (approved):
OK, let's go with this.

/ocabot merge major

pedrobaeza:
I'm seeing a hole if the resource calendar is modified for already generated entries. Have you taken this into account?

etobella:
:thinking: nice catch, maybe the right approach would be to have a wizard to recheck this information. WDYT?

pedrobaeza:
I think it's better to act automatically on the write:

- In the resource calendar.
- In the employee if the resource calendar is changed.

etobella:
I do not agree with that. Let's imagine I have a M-F 8-16 Working Calendar and I change my calender to 9-15. I don't want all history to be changed. It was correct. I know that there is resource calendar planning  but that is not a required dependency here (and I think it shouldn't be) :thinking:

pedrobaeza:
OK, you are right. Then I think it should be offered at employee and calendar level a button to refresh it selecting a date interval.

PR #248, opened 2026-01-30
[17.0][ADD] hr_attendance_geolocation_required

Neitherkx:
Supersede https://github.com/OCA/hr-attendance/pull/215
This PR updates the Attendance Geolocation Required module (hr_attendance_geolocation_required).


hr_attendance_geolocation_required/controllers/main.py
```diff
@@ -0,0 +1,19 @@
...
+
+from odoo.addons.hr_attendance.controllers.main import HrAttendance
+
+
+class HrAttendanceCustom(HrAttendance):
+    @http.route()
+    def manual_selection_with_geolocation(
+        self, token, employee_id, pin_code, latitude=False, longitude=False
+    ):
+        if not latitude or not longitude:
```
petrus-v: what about leaving at special 0.000 latitude or longitude... well I guess exact 0.00 should be very rare so seems fine to me

hr_attendance_geolocation_required/controllers/main.py
```diff
@@ -0,0 +1,19 @@
...
+            return {"error": "You must activate location and GPS to clock in."}
```
petrus-v: We should probably allow user to translate this string ?


```suggestion
            return {"error": _("You must activate location and GPS to clock in.")}
```

petrus-v (approved):
Code review looks good to me, I let two possible improvement point

hr_attendance_geolocation_required/controllers/main.py
Neitherkx: You're right, the probability of someone being exactly at 0.0, 0.0 is extremely rare. I agree the current implementation is fine for this use case.

hr_attendance_geolocation_required/controllers/main.py
Neitherkx: I've applied the change, the string is now translatable. Thanks for the feedback!

hr_attendance_geolocation_required/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2025 Álvaro Alonso Bada - Grupo Isonor
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "HR Attendance Geolocation Required",
+    "summary": "Hace obligatoria la geolocalización para fichar en Odoo 17",
```
rrebollo:
```suggestion
    "summary": "This module makes geolocation mandatory for check-in/check-out",
```
The proposed change translates the string into English (previously it was in Spanish, which was not correct). It also removes the hardcoded reference to Odoo 17, so the text will no longer need to be modified with each new Odoo version.

rrebollo (changes requested):
Great work! This is really relevant.
Would you be open to making this configurable at  the company level? Right now the module applies to all companies in the instance.
Note that the requested change is only for the Spanish summary text. Making it company-aware is just a suggestion — we can address that after merging if you'd like.

rrebollo:
@Neitherkx good work again. I saw you implemented both of my suggestions. Great!!!

Yesterday, while rethinking the implementation, I asked myself: why didn't you go for an `@api.constrains` approach, at least as a final layer of restriction at the very root level (`hr.attendance`)?

I'm guessing the JavaScript layer of restriction is sufficient for the web UI experience, but with the constraint approach, perhaps we wouldn't need the controller layer at all.

What are your thoughts on this? Am I overcomplicating things?

Neitherkx:
> @Neitherkx good work again. I saw you implemented both of my suggestions. Great!!!
> 
> Yesterday, while rethinking the implementation, I asked myself: why didn't you go for an `@api.constrains` approach, at least as a final layer of restriction at the very root level (`hr.attendance`)?
> 
> I'm guessing the JavaScript layer of restriction is sufficient for the web UI experience, but with the constraint approach, perhaps we wouldn't need the controller layer at all.
> 
> What are your thoughts on this? Am I overcomplicating things?

First of all, thank you for your feedback.

Good point. I actually implemented the defense-in-depth layer you suggested, but as an override of _attendance_action_change on hr.employee instead of @api.constrains on hr.attendance.

I chose _attendance_action_change because:

- It validates before the record is created (vs @api.constrains which triggers post-create with rollback on failure)
- It handles both check-in and check-out with a single check on geo_information, while a constraint would need to distinguish between in_latitude/in_longitude on create and out_latitude/out_longitude on write
- It catches the most realistic bypass scenario: another module calling _attendance_action_change() without geo_information

So we now have two layers:

- Controller: returns clean JSON error -> good kiosk UX
- Model (_attendance_action_change): raises ValidationError -> safety net for any code path that bypasses the controller

What do you think?

Once approved, I plan to apply these changes to my other PR in version 18.

Thanks again!

Danimarqz (approved):

rrebollo:
@Neitherkx Your approach makes complete sense to me. In fact, I’d like you to take a look at the implementation of a companion addon we’re proposing in a PR, where I actually used the API contrains approach I previously suggested to you. However, after thinking more about it, I’m starting to believe that your approach might lead to better results.

Could you please review specifically the validation logic where I’m using contrains, and also the overall implementation at the HR attendance model level, and share your perspective? #265

Neitherkx:
> @Neitherkx Your approach makes complete sense to me. In fact, I’d like you to take a look at the implementation of a companion addon we’re proposing in a PR, where I actually used the API contrains approach I previously suggested to you. However, after thinking more about it, I’m starting to believe that your approach might lead to better results.
> 
> Could you please review specifically the validation logic where I’m using contrains, and also the overall implementation at the HR attendance model level, and share your perspective? #265

I think that the @api.constrains usage here is fine. It validates that the work location belongs to the same company as the employee, these are purely relational fields that live directly on the record. 

For our geolocation requirement module, it's different, we need to check whether the browser actually sent GPS coordinates, but that information gets mixed with GeoIP fallback data in _get_geoip_response() before it even reaches hr.attendance. By the time a constraint would fire, we can't tell the difference between "real GPS from the browser" and "server-side GeoIP approximation". That's why we do our validation in _attendance_action_change instead.

rrebollo (approved):
Code Review. LGTM!

ajfebres:
Code Review. LGTM!

rrebollo:
@Neitherkx, please rebase — the Runboat link is broken.

rrebollo:
@etobella do you think this could be merge?

ajfebres:
@etobella PING

PR #256, opened 2026-03-04
[19.0][MIG] hr_attendance_reason: Migration to 19.0

EndikaPM:


hr_attendance_reason/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     "author": "Odoo S.A., Tecnativa, Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "installable": True,
-    "depends": ["hr_attendance"],
+    "depends": ["hr_attendance", "mail"],
```
MiquelRForgeFlow, line 13: no need, already depends on `mail` indirectly.

hr_attendance_reason/models/hr_attendance.py
```diff
@@ -7,10 +7,16 @@
 
 class HrAttendance(models.Model):
     _inherit = "hr.attendance"
+    _mro_pad = ["mail.thread"]
```
MiquelRForgeFlow, line 10: what is this? the inherit is already in odoo, see https://github.com/odoo/odoo/blob/19.0/addons/hr_attendance/models/hr_attendance.py#L31

hr_attendance_reason/views/hr_attendance_view.xml
```diff
@@ -17,6 +17,9 @@
                     </group>
                 </group>
             </sheet>
+            <div class="oe_chatter" position="inside">
+                <chatter />
+            </div>
```
MiquelRForgeFlow, line 22: Wrong. See https://github.com/odoo/odoo/blob/19.0/addons/hr_attendance/views/hr_attendance_view.xml#L232.

MiquelRForgeFlow (changes requested):

victoralmau:
Superseed by https://github.com/OCA/hr-attendance/pull/315

PR #258, opened 2026-03-05
[15.0][FIX] hr_attendance_rfid: if there's an error do not register the attendance

AaronHForgeFlow:
Currently the try except makes the attendance to be created when there is an exception, tested with errors raised on the _check_attendance method from Odoo module hr_attendance

@lubusax


MiquelRForgeFlow (approved):

GuillemCForgeFlow (approved):
lgtm 👍🏿

PR #264, opened 2026-03-23
[18.0][ADD] hr_attendance_manage_own

AungKoKoLin1997:
This module allows users to create and edit their own attendance records without requiring attendance manager or officer privileges, with configurable field-level restrictions.

Without this module, regular internal users can only read their own attendance records. Creating or modifying attendance records requires either the Attendance Manager role (full access to all records) or the Attendance Officer role with the employee's attendance manager set to that user.

This module introduces an "Own Manager" group that can create and edit attendance records, except for fields configured as disallowed. The `overtime_status` field is disallowed by default. Additionally, when creating records, the `overtime_status` must be set to 'To Approve', and attendance records that have been approved or refused cannot be modified.

@qrtl QT6380


kanda999 (approved):
Functional and code review.

yostashiro (commented):
Haven't reviewed the code yet, but a suggestion on the module name: hr_attendance_manage_own

I think this better fits the OCA naming convention than the current one.

hr_attendance_manage_own/models/hr_attendance_disallowed_field.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class HrAttendanceDisallowedField(models.Model):
```
yostashiro: How about adding a many2many `attendance_restricted_field_ids` field to res.company instead of adding a new model? I think that would make the implementation a bit simpler and more flexible.

hr_attendance_manage_own/security/hr_attendance_manage_own_security.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="group_hr_attendance_own_manager" model="res.groups">
+        <field name="name">Own Manager: Manage own attendance records</field>
```
yostashiro:
```suggestion
        <field name="name">User: Manage own attendances</field>
```

hr_attendance_manage_own/models/hr_attendance_disallowed_field.py
AungKoKoLin1997: Yes. I update even I have concern about admin can only update the fields list and every attendance manager may not have admin access right. But I think that is enough for the operation.

hr_attendance_manage_own/security/hr_attendance_manage_own_security.xml
AungKoKoLin1997: Done!

yostashiro (commented):
@AungKoKoLin1997 I've improved the design. Can you please review my commit?

Main points:
- removed the restricted field config since there probably aren't many cases where fields other than overtime_status and validated_overtime_hours should be protected, and applying the same check to create and write wouldn't always work
- added support for delete operations

AungKoKoLin1997:
> I've improved the design. Can you please review my commit?

@yostashiro I reviewed the code and functional. It looks good.

yostashiro:
@AungKoKoLin1997 Can you please rebase and squash commits.

AungKoKoLin1997:
> Can you please rebase and squash commits.

@yostashiro Done! I also added you as a contributor.

yostashiro (approved):
Thanks! LGTM.

SanamiK:
I applied this improvement:
Add a configurable system parameter (hr_attendance_manage_own.write_bypass_fields) that allows specified fields to be written on processed (approved/refused) attendance records without triggering the access error. This is useful when other modules (e.g. hr_timesheet_sheet) need to link attendance records via fields like sheet_id.

@qrtl QT6692

SanamiK (approved):

PR #251, opened 2026-02-09
[FIX] hr_attendance_autoclose: launch test at install

petrus-v:
I've spend that much of time to try to figure out what changed, make the test stable.

(sorry, couldn't find how to change the PR name afterwards)

PR #265, opened 2026-03-26
[17.0][ADD] hr_attendance_work_location

rrebollo:
This pull request introduces the `hr_attendance_work_location` module, which
enhances the HR Attendance app by enabling work location tracking for employee
attendance records.

**Features:**

-   **Automatic mode (GPS):** Automatically assigns work locations at check-in
    and check-out based on GPS coordinates, with configurable proximity tolerance
    (set per company in **HR Settings > Attendance**, default: 111 meters).

-   **Manual mode (Selector):** When set to manual mode, the employee selects
    their work location before each check-in and check-out. This works in two
    interfaces:
    -   **Kiosk mode:** After employee identification (barcode scan, manual
        selection, or PIN), an intermediate screen appears where the employee
        selects their work location. The location is shown read-only on the
        greeting screen after the operation.
    -   **Backend systray:** The work location dropdown is always visible in
        the attendance popup and editable before both check-in and check-out.
        The selected location is stored in the **Check-in Work Location** or
        **Check-out Work Location** field on the attendance record accordingly.

-   **Configurable per company:** Choose between automatic and manual modes,
    set a default work location, and optionally require location selection.

-   **Exclude from attendance:** Work locations marked as excluded are never
    assigned automatically but remain available for manual selection.


hitrosol:
Nice module @rrebollo, 
Would it be acceptable to have configurable of the GEO_TOLERANCE_DEGREES ? To my knowledge different companies have different geo tolerance degrees,

rrebollo:
> Nice module @rrebollo, Would it be acceptable to have configurable of the GEO_TOLERANCE_DEGREES ? To my knowledge different companies have different geo tolerance degrees,

Do you mean configurable through the web UI—perhaps as a system parameter that applies to all companies in the same instance—or at the individual company level?

I would prefer the system parameter implementation; I think it's more practical. Do you have any use cases that would require the company-level approach?

hitrosol:
Currently, there’s no specific use case for it, but I suggested it to anticipate future needs and more flexible.
By the way, will this implementation be compatible with the Required Geolocation module? Or it can be as one module to make sure the geolocation is filled correctly.

rrebollo:
> By the way, will this implementation be compatible with the Required Geolocation module? Or it can be as one module to make sure the geolocation is filled correctly.

We designed it without any coupling. It is the implementer's responsibility to ensure geolocation data is provided, but how that is done falls outside the scope of this addon.

In fact, right now we are using the addon you mentioned.

rrebollo:
~~Some hr_attendance_autoclose test was failing. No it's fixed and this PR now depends on it #251.~~

rrebollo:
@hitrosol, the geotolerance feature is now configurable at the company level — a value of zero will disable it. Can you give me your review?

rrebollo:
**Update:** After a team member raised a question about how to configure the tolerance for 50 meters, we realized the field was expressed in decimal degrees — not intuitive for end users. We've updated the module so the tolerance is now configured in **meters** (default: 111m). Internally, the algorithm converts to degrees at runtime.

Changes included:
- Renamed `geo_tolerance_degrees` → `geo_tolerance_meters` on `res.company`
- Added meter-to-degree conversion in `_compute_work_locations()`
- Updated documentation (`CONFIGURE.md`) with a reference table for common values and device accuracy guidance
- Updated all tests and Spanish translations

PR #267, opened 2026-04-05
[17.0] [IX] hr_attendance_autoclose: fix test_employee_edit expects wrong worked_hours

rrebollo:
### Context

TestHrAttendanceReason.test_employee_edit fails with 11.0 != 10.0. The test creates an attendance 15 hours ago with no check_out, triggers autoclose (which sets check_out = check_in + 11h), then asserts worked_hours == 10.0 — assuming Odoo core deducts 1 hour for lunch.

The lunch deduction in _get_worked_hours_in_range (core hr_attendance/models/hr_attendance.py:186) depends on      _employee_attendance_intervals(..., lunch=True) returning lunch intervals from the employee's resource calendar.

The test employee is created with no explicit calendar ({"name": "Employee"}), and the company's default calendar in the test environment no longer provides a lunch break interval — so worked_hours = 11.0 (raw delta, no deduction).

The test's comment on line 41-42 references Odoo commit 2eda54348de1 for the lunch logic, but this is fragile: the test is asserting on Odoo core's lunch behaviour which is outside this module's scope.

### Fix
hr_attendance_autoclose/tests/test_hr_attendance_auto_close.py

Line 43: Change the assertion to not depend on lunch deduction. The autoclose module's job is to set check_out = check_in + max_hours (default 11). Assert on what the module controls:
     - Assert att.check_out is set (autoclose worked)
     - Assert att.worked_hours matches the actual raw delta without assuming lunch deduction

Simplest fix: change 10.0 → 11.0 and update the comment to reflect that lunch deduction depends on calendar config and is not this module's concern.

### DIFF
--- a/hr_attendance_autoclose/tests/test_hr_attendance_auto_close.py
+++ b/hr_attendance_autoclose/tests/test_hr_attendance_auto_close.py
@@ -38,9 +38,10 @@ class TestHrAttendanceReason(BaseCommon):
             {"employee_id": self.employee.id, "check_in": dt.strftime(DF)}
         )
         self.hr_attendance.check_for_incomplete_attendances()
-        # worked_hours are now 10 hours, because Odoo adds 1 hour to lunch, see:
-        # https://github.com/odoo/odoo/commit/2eda54348de1bd42fc2a1bed94cd8b7a3ebf405d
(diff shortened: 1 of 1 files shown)


rrebollo:
Closing this in favor of the much more robust approach in #251.

PR #268, opened 2026-04-10
[18.0][FIX] hr_attendance_reason: fix argument attendance_reason_id

jans23:


hr_attendance_reason/controllers/main.py
```diff
@@ -32,21 +32,25 @@ def user_attendance_data(self):
...
-        if request.params.get("attendance_reason_id"):
-            request.update_context(
-                attendance_reason_id=int(request.params.get("attendance_reason_id"))
-            )
+    def systray_attendance(
+        self, latitude=False, longitude=False, attendance_reason_id=False
+    ):
+        if attendance_reason_id:
+            request.update_context(attendance_reason_id=int(attendance_reason_id))
         return super().systray_attendance(latitude=latitude, longitude=longitude)
```
victoralmau, line 40: I don't understand this change based on the existing method in `hr_attendance` https://github.com/odoo/odoo/blob/d73e865876c0c1d1124b32e9dea2ef7b73b667dc/addons/hr_attendance/controllers/main.py#L194

hr_attendance_reason/controllers/main.py
```diff
@@ -32,21 +32,25 @@ def user_attendance_data(self):
...
+        latitude=False,
+        longitude=False,
+        attendance_reason_id=False,
     ):
-        if request.params.get("attendance_reason_id"):
-            request.update_context(
-                attendance_reason_id=int(request.params.get("attendance_reason_id"))
-            )
+        if attendance_reason_id:
+            request.update_context(attendance_reason_id=int(attendance_reason_id))
```
victoralmau, line 53: I don't understand this change based on the existing method in `hr_attendance` https://github.com/odoo/odoo/blob/d73e865876c0c1d1124b32e9dea2ef7b73b667dc/addons/hr_attendance/controllers/main.py#L158

victoralmau (changes requested):

jans23:
Thank you for your feedback. We checked this change again and indeed it seems to be unreasonable.

PR #269, opened 2026-04-13
[17.0][FIX] hr_attendance_autoclose: fix flaky worked_hours test

BhaveshHeliconia:
This PR addresses an intermittent failure in the `hr_attendance_autoclose` module affecting the `test_employee_edit` test, which could raise:

> `AssertionError: 11.0 != 10.0`

### **Cause of the Issue**

The failure is caused by how Odoo 17 computes the `worked_hours` field, which depends on timezone-aware lunch break deductions:

* **Timezone impact:** In timezones like IST, UTC-based check-in times may shift into or out of the lunch break window (12:00–13:00), resulting in inconsistent deduction of the 1-hour break.
* **Non-deterministic dates:** The use of `datetime.now()` made the test sensitive to the execution day. For example, running the test on a Monday could place the reference check-in on Sunday, where no lunch break applies.

### **Fix Implemented**

- Timezone Synchronization: Forced the test Company, Resource Calendar, and Employee resources to use UTC in setUpClass. This guarantees that UTC-based test data consistently overlaps with the calendar's 12:00–13:00 lunch break deduction window.

- Dynamic Workday Logic: Replaced hardcoded/current dates with a dynamic calculation that always targets the previous Wednesday at 09:00 AM. This ensures the test always runs on a standard working day with active lunch segments, without relying on the current wall clock.

### DIFF
--- a/hr_attendance_autoclose/tests/test_hr_attendance_auto_close.py
+++ b/hr_attendance_autoclose/tests/test_hr_attendance_auto_close.py
@@ -18,11 +18,21 @@ class TestHrAttendanceReason(BaseCommon):
         super().setUpClass()
         cls.hr_attendance = cls.env["hr.attendance"]
         cls.employee = cls.env["hr.employee"].create({"name": "Employee"})
+        # Force UTC to ensure deterministic lunch deduction (09:00-20:00 UTC includes
+        # 12:00-13:00)
+        cls.env.company.resource_calendar_id.write({"tz": "UTC"})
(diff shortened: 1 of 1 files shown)


Tardo (approved):

BhaveshHeliconia:
ping @etobella

BhaveshHeliconia:
@etobella Could you please review and merge this? Thanks!

etobella (approved):
Makes sense, thanks for forcing up and leave less space to timezone problems 😉

PR #271, opened 2026-04-15
[ADD] hr_attendance_rfid_log: new module for attendances logging

lubusax:
This new module enhances the HR Attendance functionalities by integrating a logging system of all card readings coming from RFID card readers linked to the hr attendance capabilities of Odoo. It captures, stores, and manages attendance events generated by RFID devices, providing an intermediary layer that ensures reliability and comprehensive tracking before the attendance is recorded in the Odoo attendances database.

## Needs Addressed

In many organizations, attendance tracking using hardware devices (like RFID readers) faces challenges such as network disconnections, unregistered cards, missing clocking entries (that cause errors in the Odoo logic) or delayed synchronizations. This module covers these needs by:

1.  **Robust Event Logging:** Instead of directly modifying the HR attendance records, the module logs all incoming RFID events into a dedicated model (`hr.attendance.rfid.log`).
2.  **State Management:** Every scan gets categorized into actionable states:
    *   **Success:** The RFID scan was successfully linked to an employee and registered as a valid check-in or check-out.
    *   **Failed:** An error occurred (e.g., the RFID card is not associated with any employee or the attendance logic rules were violated).
    *   **Retry:** Events that previously failed but are marked to be retried (e.g., after the card has been assigned to a user).
    *   **Ignore:** Events that administrators actively choose to dismiss, removing them from the queue of pending issues.
3.  **Traceability:** It maintains historical logs of when the event happened, the card code used, the associated IoT device, and any error messages generated during the process.


AaronHForgeFlow:
Hi @OCA/human-resources-maintainers could you please allow the workflow to run? thank you.

hr_attendance_rfid_log/models/hr_employee.py
```diff
@@ -0,0 +1,45 @@
...
+            "error_message": res.get("error_message"),
+            "logged": res.get("logged"),
+            "action": res.get("action"),
+            "timestamp": fields.Datetime.now(),
+        }
+        return vals
+
+    @api.model
+    def register_attendance_with_log(self, log):
+        with freezegun.freeze_time(
```
AaronHForgeFlow: Hi @lubusax, this is a strong dependency and can cause some side-effects, so I wonder if we can remove it or  perhaps modify the register_attendance method to accept an optional timestamp parameter and pass that through to the check_in/check_out fields of the attendance record.

hr_attendance_rfid_log/models/hr_attendance_rfid_log.py
```diff
@@ -0,0 +1,124 @@
...
+                if purge_period_success
+                else default_purge_period_success
+            )
+            purge_period_general = (
+                int(purge_period_general)
+                if purge_period_general
+                else default_purge_period_general
+            )
+        except ValueError:
+            self.env["ir.config_parameter"].sudo().set_param(
```
AaronHForgeFlow: Writing a config parameter because of a fail is not conventional. I think maybe we can use a company parameter with a default value to avoid users putting a wrong value in this parameter

hr_attendance_rfid_log/models/hr_attendance_rfid_log.py
```diff
@@ -0,0 +1,124 @@
...
+    employee_id = fields.Many2one(
+        "hr.employee", string="Employee", ondelete="cascade", index=True
+    )
+    error_message = fields.Char(string="Error Message", readonly=True)
+    logged = fields.Boolean(string="Logged", readonly=True)
+    action = fields.Selection(
+        selection=[
+            ("check_in", "check in"),
+            ("check_out", "check out"),
+            ("FALSE", "Not Defined (in or out)"),
```
AaronHForgeFlow: False can be confusing, perhaps we can use "undefined"

hr_attendance_rfid_log/models/hr_attendance_rfid_log.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def action_set_to_retry_state(self):
+        self.sudo().state = "retry"
+
+    def action_set_to_ignore_state(self):
+        self.sudo().state = "ignore"
+
+    def action_retry_now(self):
+        if self.state == "retry":
+            self.sudo().retry_counter += 1
```
AaronHForgeFlow: It seems using sudo() here can be not efficient, if really needed we can use it in the _retry_attendance_rfid_log before the loop

hr_attendance_rfid_log/models/hr_attendance_rfid_log.py
```diff
@@ -0,0 +1,124 @@
...
+
+    state = fields.Selection(
+        selection=[
+            ("success", "Success"),
+            ("failed", "Failed"),
+            ("retry", "Retry"),
+            ("ignore", "Ignore"),
+        ],
+        string="State",
+        readonly=True,
```
AaronHForgeFlow, line 22: perhaps we can put an index=True

AaronHForgeFlow (commented):
Hi @lubusax, thank you for the contribution. I made some comments. Can you please check if they make sense to you? Thanks.

hr_attendance_rfid_log/models/hr_employee.py
```diff
@@ -0,0 +1,45 @@
...
+            datetime.fromtimestamp(log.timestamp.timestamp(), tz=None)
+        ):
+            result = self.register_attendance(log.rfid_card_code, log,)
```
etobella: This has nothing to do with the goal of the module...

This shouldn't be here and looks like a customization you need, however, it shouldn't be here.

lubusax:
Hi @AaronHForgeFlow and @etobella,

Thank you for the detailed review and guidance! I have pushed an update that addresses all the points raised:

Removed freezegun: I removed this dependency entirely. Instead, I updated the register_attendance signature defensively (timestamp=False, kwargs) and used context injection to pass the custom time down to an override on hr.attendance's create and write methods. This safely applies the timestamp without duplicating core logic or risking substitution errors.

Config Parameter Fallback: Removed the set_param write in the except block. It now simply falls back to the default variables in memory if a ValueError occurs.

Action State Label: I updated the UI display label to "undefined" as suggested.

Loop Efficiency: Moved the sudo() elevation outside of the loop in _retry_attendance_rfid_log to prevent redundant queries.

Indexes: Added index=True to both the state and employee_id fields for better database performance during cron executions.

Let me know if the approach aligns with your expectations.

hr_attendance_rfid_log/models/hr_attendance.py
```diff
@@ -0,0 +1,33 @@
...
+
+from odoo import api, models
+
+
+class HrAttendance(models.Model):
+    _inherit = "hr.attendance"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """
```
etobella, line 12: Same comment applies...

This is your customization. that shouldn't be here. make a custom module for this thing, don't use OCA for that.

etobella (changes requested):
You are adding unnecessary things for your own customization. Please stop.

Also, avoid AI answers please...

PR #274, opened 2026-04-23
[16.0][FIX] hr_contract_update_overtime: make leave no allocation required + [UPD] copier templates

luisDIXMIT:
When holiday_id is included in the leave fields, this occurs when the hr_holidays Odoo module is installed, the setup method creates a leave for a newly created leave type, but this leave type requires allocation by default. Therefore, if we want to create a leave without allocation, we must configure the leave type so that no allocation is required.


luisDIXMIT:
We can see how the previous workflow fails in this PR #247

luisDIXMIT:
Can I merge this PR into #275 to have a green CI? @etobella @pedrobaeza I will keep separated commits.

etobella:
Include the other commit here. We need both commis together, so let's keep them together :wink:

etobella (approved):

AaronHForgeFlow (approved):
👍  code LG

PR #281, opened 2026-06-12
[16.0][ADD] hr_attendance_report_pdf: addind attendance pdf report

ajfebres:
@BinhexTeam

This module adds a PDF attendance report wizard to Odoo. It allows selecting date ranges, employees or departments, and choosing between summary or detailed mode. It includes options to show empty days, include open attendance entries, and choose the hours format (HH:MM or decimals). It is designed to produce clear employee or department reports with totals and subtotals useful for reviews and audits.


Neitherkx:
Hello @ajfebres, could you please take a look at this PR https://github.com/OCA/hr-attendance/pull/242? It is related to hr_attendance_report.

ajfebres:
PING @rrebollo

rrebollo:
At this point, I think the best approach is to align with the other comments already made here on the PR. Since the other PR is older and appears to share common ground with the ideas behind both addons—curiously, even sharing names—there is a clear naming collision.

We should find a strategy to resolve this. I believe the best way forward would be to unify efforts, perhaps by aligning both implementations under a common purpose. This could be done by making suggestions or contributing directly to the other PR, even though it targets a different version (17.0), and then backporting those changes here to version 16.0.

The alternative would be to contribute further if the only real issue turns out to be the naming conflict—but only the authors can assess whether there's enough common territory. If not, both should justify their approach. In this particular PR (the newer one), the author should clearly justify why a separate integration, test, and addon are needed, and explain how this addon differs from the one in the other PR despite the version difference.

Otherwise, this could cause significant confusion in the community. Someone looking for the addon by name might assume the 16.0 version is the same as the 17.0 version, leading to all kinds of misunderstandings.

hr_attendance_report_pdf/report/hr_attendance_report.py
```diff
@@ -0,0 +1,17 @@
...
+    def _get_report_values(self, docids, data=None):
+        service = self.env["hr.attendance.report.service"]
+        report_data = service._prepare_report_values(data)
+
+        return {
+            "doc_ids": docids,
+            "doc_model": "hr.attendance",
+            **report_data,
+            "tz": self.env.user.tz,
+        }
```
rrebollo, line 17: The right way is to call `super()`, so you don't need to manually set things like `doc_is` and similar fields.

hr_attendance_report_pdf/report/hr_attendance_report_service.py
```diff
@@ -0,0 +1,270 @@
+import copy
+from collections import Counter
+from datetime import datetime, timedelta
+
+from odoo import _, fields, models
```
rrebollo, line 5:
```suggestion
from odoo import _, api, fields, models
```

hr_attendance_report_pdf/report/hr_attendance_report_service.py
```diff
@@ -0,0 +1,270 @@
...
+
+        totals = self._get_totals(data, lines, docs)
+        return {
+            "docs": docs,
+            "lines": lines,
+            "totals": totals,
+            "datas": data,
+        }
+
+    def _normalize_data(self, data):
```
rrebollo, line 28:
```suggestion
    @api.model
    def _normalize_data(self, data):
```

rrebollo: I think the other methods here don't need a recordset neither.

hr_attendance_report_pdf/report/hr_attendance_report_service.py
```diff
@@ -0,0 +1,270 @@
...
+                "worked_hours": (end - check_in).total_seconds() / 3600,
+            },
+            next_day.date(): {
+                "check_in": next_day,
+                "check_out": check_out,
+                "worked_hours": (check_out - next_day).total_seconds() / 3600,
+            },
+        }
+
+    def _get_attendance_by_employees(self, data):
```
rrebollo, line 109: This method is quite long. I think splitting it into smaller pieces would improve maintainability and readability.

hr_attendance_report_pdf/report/hr_attendance_report_service.py
```diff
@@ -0,0 +1,270 @@
...
+            data["date_to"],
+            data["date_from"],
+            data["date_to"],
+        ]
+        if data["company_id"]:
+            query = query.replace("WHERE", "WHERE emp.company_id = %s AND")
+            params.insert(0, data["company_id"].id)
+
+        self.env.cr.execute(query, tuple(params))
+        ids = [row[0] for row in self._cr.fetchall()]
```
rrebollo, line 133:
```suggestion
        ids = [row[0] for row in self.env.cr.fetchall()]
```
For the sake of consistency.

hr_attendance_report_pdf/report/hr_attendance_report_template.xml
rrebollo, line 1: I would use more Odoo‑bootstrap classes instead of inline styles to maintain visual consistency.

hr_attendance_report_pdf/report/hr_attendance_report_template.xml
```diff
@@ -0,0 +1,311 @@
...
+
+                        <t t-call="hr_attendance_report_pdf.report_attendance_totals">
+                            <t t-set="department" t-value="department" />
+                        </t>
+                        <div
+                            t-if="department != list(lines.keys())[-1]"
+                            style="page-break-after: always;"
+                        />
+                    </t>
+                    <!-- <t t-call="hr_attendance_report_pdf.report_attendance_totals" /> -->
```
rrebollo, line 305:
```suggestion
```

hr_attendance_report_pdf/wizard/hr_attendance_report_wizard.py
```diff
@@ -0,0 +1,102 @@
...
+        if self.date_from and self.date_to and self.date_from > self.date_to:
+            raise ValidationError(_("Start date cannot be after end date."))
+        if self.report_type == "individual" and not self.employee_ids:
+            raise ValidationError(
+                _("You must select an employee for the individual report.")
+            )
+        if self.report_type == "department" and not self.department_ids:
+            raise ValidationError(
+                _("You must select a department for the department report.")
+            )
```
rrebollo, line 82: Please use framework helpers like @api.constrains

rrebollo (changes requested):
Please consider my suggestions

hr_attendance_report_pdf/report/hr_attendance_report.py
ajfebres: To new instances of  _AbstractModel_, I can´t super().

> File "/opt/odoo/auto/addons/hr_attendance_report_pdf/report/hr_attendance_report.py", line 13, in _get_report_values
    return super()._get_report_values(docids, data)
AttributeError: 'super' object has no attribute '_get_report_values'

PR #285, opened 2026-07-07
[18.0][MIG] hr_attendance_geolocation:  Migration to 18.0

ajfebres:
@BinhexTeam

Standard migration


rrebollo:
Given that similar features appear to be built into Odoo since version 17 (as mentioned in [this comment](https://github.com/OCA/hr-attendance/issues/154#issuecomment-3593868262)), could you clarify what additional functionality or benefit this PR provides that makes it worth merging? Is it addressing a gap that the standard features do not cover?

PR #288, opened 2026-07-15
[19.0][ADD] hr_attendance_break

bosd:
New module: record employee **breaks** during attendance (clocked-in) time, for
labour-compliance reporting (are breaks too long?).

## Features
- New `hr.attendance.break` model — one record per break (start/stop/type), so a
  full day can hold several breaks (morning coffee, lunch, afternoon coffee).
- `hr.attendance` gains `break_hours` and `net_worked_hours` (worked minus
  breaks), plus a **Breaks** reporting menu (list + pivot).
- **Start / End Break** button in the top-bar attendance (systray) menu — always
  available while checked in, regardless of the work being done — via a
  `/hr_attendance/toggle_break` route. Open breaks auto-close on check-out.
- **Attendance kiosk**: a checked-in employee who selects their name is offered
  *Start/End Break* or *Check out* (public_kiosk OWL patch + pin-checked routes).
- Configurable **break rounding** (`res.company` / Settings, `0` = off).

## Tests
5/5 tests pass on 19.0 (toggle create/close, multi-break sum, net worked hours,
auto-close on check-out, rounding, check-in guard).

## Notes
First 19.0 module in this repo. The kiosk OWL patch mirrors `hr_attendance_reason`;
it has not been browser/tour-tested and should be verified against the target
deployment's Odoo minor version. The systray button is the primary, version-stable
surface.


pedrobaeza:
Did you know https://github.com/OCA/hr-attendance/tree/15.0/hr_attendance_rest_time_included ?

bosd:
Thanks @pedrobaeza. Did'nt know that module. I went through hr_attendance_rest_time_included (it's on 15.0 and 18.0). There's genuine overlap in the data both keep multiple breaks per attendance . but the capture and semantics differ: it's reason-driven (check out/in with a rest_time_included reason, which reopens the attendance) and it includes rest in worked hours, whereas some of our agreements need an unpaid/deducted break, and several clients don't want to pick a reason per break.

Rather than a parallel module, I think this consolidates cleanly into layers:

- `hr_attendance_break`
(base, depends hr_attendance): one lean hr.attendance.break model + break types with a paid flag, break_hours/net_worked_hours (worked hours never modified), and a live Start/End Break button in the systray + kiosk. All-paid reproduces the current "included" behaviour; lunch-unpaid gives the compliance case.
- `hr_attendance_break_reason`
(optional, + hr_attendance_reason): the creates_break reason flag and the reason-driven capture — essentially rest_time_included's flow refactored to write into the shared model. Credit to Tecnativa for this part.
- `hr_timesheet_attendance_break`
(later, + hr_timesheet_time_control): billable/timer behaviour, defaulting to not pausing the project timer.

That keeps reason, paid/unpaid, and timer-pausing all optional. Would Tecnativa/you be open to converging on a shared base like this (I'm happy to do the refactor and keep rest_time_included's behaviour as the reason layer), or do you prefer a different split? I'll follow whatever the repo maintainers favour.

pedrobaeza:
Hi, @bosd for us it's important to unify the buttons, and not to have a separate one for doing the breaks. It's a source of mistakes. That's why we used the reasons to determine that, but using the same check-in/check-out buttons, so that's a requirement that we can't bypass.

bosd:
Thanks, that's clear — I'll drop the separate break button and align on the shared check-in/check-out + reasons flow. It actually fits our need: clients just want a deliberate tap to **start** and to **end** a break, and that maps straight onto *check-out (with a break reason) → check-in*, so no extra button is required.

On the reason-friction worry I raised earlier: I dug into the code and it's largely a non-issue by default. `show_reason_on_attendance_screen` and `required_reason_on_attendance_screen` are both **off by default**, and the requirement is only enforced client-side when `required` is on. So a clean minimal-friction setup is `show = on`, `required = off`: an employee picks a reason **only when starting a break** (one tap), and needs **no reason to check out** for the day. A configurable **default break reason** would make even that near-zero friction — would you be open to that?

The part I'd still like to contribute is orthogonal to the trigger: a lean base module `hr_attendance_break` with **break types + a per-type paid/unpaid flag**, `break_hours`/`net_worked_hours` (worked hours never modified), and reporting. All-paid reproduces today's "included" behaviour; unpaid lunch covers agreements where breaks aren't paid — which `rest_time_included` doesn't handle yet. `rest_time_included` would then store its breaks in that shared model, keeping your reason-based, single-button capture exactly as-is.

One practical note: neither `hr_attendance_reason` nor `rest_time_included` is on **19.0** yet (they stop at 18.0), so converging there means porting both first — I'm happy to take that on. Does that direction, and the default-reason idea, work for you?

PR #289, opened 2026-07-17
[19.0][MIG] hr_attendance_reason: Migration to 19.0

farhatzafar:
Migration of `hr_attendance_reason` to 19.0.

Changes include:
- Bumped manifest version to `19.0.1.0.0`
- Replaced `_sql_constraints` with `models.Constraint`
- Updated controller routes from `type="json"` to `type="jsonrpc"`
- Adapted kiosk manual selection override to the v19 `manual_selection` method
- Adjusted attendance systray flow for v19 response behavior
- Added lazy session enrichment so reasons are available on first systray render
- Adjusted attendance configuration view/menu behavior for v19

Manual tests performed:
- Settings block and Reasons menu
- Reason creation and default sign-in/sign-out reasons
- Systray check-in/check-out with reasons
- Required reason validation
- Kiosk manual selection with reasons
- Reason saving on attendance records
- Compared behavior against the original 18.0 module


hr_attendance_reason/security/security.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="rule_multi_company_hr_attendance_reason" model="ir.rule">
+        <field name="name">Attendance Reason multi-company</field>
+        <field name="model_id" ref="model_hr_attendance_reason" />
+        <field name="global" eval="True" />
+        <field
+            name="domain_force"
+        >['|',('company_id','=',False),('company_id','in',company_ids)]</field>
```
victoralmau, line 9:
```suggestion
        > [('company_id', 'in', company_ids + [False])]</field>
```

victoralmau (changes requested):
Please, squash migration commits.

Image

haivanlive:
In Odoo 19, standard employees no longer have access to the Attendance module directly. Instead, they must view their attendance records through the Employee module.

Because of this change, we should adjust this module so that employees can view and add attendance reasons directly from the attendance view/tab inside the Employee module.

haivanlive:
When the reason is missing, could we display a more friendly/polite error message instead?
image

haivanlive:
When I select a reason during check-in / check-out, the reason is not being saved to the attendance record.
image

victoralmau:
Superseed by https://github.com/OCA/hr-attendance/pull/315

PR #290, opened 2026-07-24
[18.0][IMP] hr_attendance_report_theoretical_time: Add leave hours column to report 

victoralmau:
Add leave hours column to report 

The Leave column is added to the report so that leave data can also be viewed for informational purposes

@Tecnativa TT62028


victoralmau:
It's ready for review. It has been tested with a customer.

carlos-lopez-tecnativa (approved):
LGTM

hr_attendance_report_theoretical_time/migrations/18.0.2.1.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
```
pedrobaeza: If you use `openupgradelib`, you can use `add_columns` with the default parameter to set it in one shot (it's done via default, which is more efficient).

hr_attendance_report_theoretical_time/i18n/es.po
```diff
@@ -162,6 +162,15 @@ msgstr "Desde"
 msgid "ID"
 msgstr "ID"
 
+#. module: hr_attendance_report_theoretical_time
+#: model:ir.model.fields,help:hr_attendance_report_theoretical_time.field_hr_leave_type__include_in_leave_theoretical
+msgid ""
+"If you check this mark, leaves in this category won't reduce the number of "
```
pedrobaeza, line 168: Just for reference, changing translations in a merged module may conflict with Weblate, but I'll check before merging.

victoralmau:
Ping @pedrobaeza

hr_attendance_report_theoretical_time/i18n/es.po
victoralmau: Excuse me, are you saying it’s not possible to add the corresponding translations—the ones added with a PR—to the PR itself? We’re not changing existing translations; we’re adding new translations that Weblate will automatically add **after** the PR is merged.

Let’s say this PR is added for a client; the relevant translations wouldn’t exist yet. How would we solve that problem?

pedrobaeza: If you add them, there's a risk to get a conflict, yes, and this is how git works, getting the previous and the following lines of the diff as reference. Imagine the previous string to yours is not translated, and it's translated meanwhile. The diff won't resolve correctly. Or if other PR is merged adding a new string before this one...

I always check before merging for avoiding this situation, but just FYI.

pedrobaeza (approved):
/ocabot merge nobump

PR #292, opened 2026-08-07
[18.0][IMP] hr_attendance_report_theoretical_time: Delete unnecessary records if the theoretical_hours_start_date field is modified

victoralmau:
Delete unnecessary records if the `theoretical_hours_start_date` field is modified

Please @pedrobaeza can you review it?

@Tecnativa


pedrobaeza (commented):
This should be in both ways: you need to create missing ones if the date is prior to the previous one.

victoralmau:
> This should be in both ways: you need to create missing ones if the date is prior to the previous one.

Done.

Another important point: there is no end date for theoretical hours. Until now, it was normal for there not to be one, but now the system will generate attendance active=false until the employee is archived.

What happens if the module is installed when there are already archived employees? The system will not generate attendance active=false, and we also won't have an employee archive date or an end date for theoretical hours. 
Any suggestions about this?

hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -13,3 +17,77 @@ class HrEmployee(models.Model):
...
+
+    def _action_create_empty_attendance(self, date_from=False, date_to=False):
+        """Method for creating inactive attendance records with a
+        duration of 0.
+        The goal is to prevent the report from having to calculate data
+        directly, so employee attendance records are created for past days
+        when they should have clocked in."""
+        self.ensure_one()
+        sql = """
+        SELECT DISTINCT(check_in)::date
```
pedrobaeza: One side question: why doing this with SQL? I think we should avoid it.

pedrobaeza:
About the active thing: that things should be done manually, as it's not the usual one to retroactively do this IMO.

hr_attendance_report_theoretical_time/models/hr_employee.py
victoralmau: Simply because that's how it was before (https://github.com/OCA/hr-attendance/pull/247); yes, I also think it could be changed.

pedrobaeza: Then we can improve it now that the code is being moved.

victoralmau:
> About the active thing: that things should be done manually, as it's not the usual one to retroactively do this IMO.

I think this is a “_common_” case: there is an archived employee (someone who has worked, for example, part of the current year), but no historical records will have been generated for them (the cron job won't generate them). Even if we did generate them, we wouldn't know the date up to which they should be generated.

pedrobaeza:
I think with contract, we can check contract dates for that, but not without it.

pedrobaeza (approved):

victoralmau:
Ping @carlos-lopez-tecnativa

hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -13,3 +17,79 @@ class HrEmployee(models.Model):
...
+    def write(self, vals):
+        res = super().write(vals)
+        # If the theoretical_hours_start_date value is modified and there are attendance
+        # records with active=false generated by the cron job, you must delete them to
+        # prevent inconsistent data from appearing in the report.
+        # We also create any pending attendance that have not been generated
+        if "theoretical_hours_start_date" in vals:
+            attendance_model = self.env["hr.attendance"].with_context(active_test=False)
+            for item in self:
+                date_from = item.theoretical_hours_start_date or item.create_date.date()
```
carlos-lopez-tecnativa: When the field is a datetime and you only need the date part, take into account that this could return the next day. In countries with a negative time zone offset, this is a common issue.
```suggestion
                date_from = item.theoretical_hours_start_date or fields.Datetime.context_timestamp(item.create_date).date()
```

hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -13,3 +17,79 @@ class HrEmployee(models.Model):
...
+                attendance_model.sudo().search(
+                    [
+                        ("employee_id", "=", item.id),
+                        ("active", "=", False),
+                        ("check_in", "<", date_from),
+                    ]
+                ).unlink()
+                item._action_create_empty_attendance(date_from, fields.Date.today())
```
carlos-lopez-tecnativa: The same applies here. I would prefer using `context_today` instead of `today` for the same reason mentioned above.
```suggestion
                item._action_create_empty_attendance(date_from, fields.Date.context_today(self))
```

carlos-lopez-tecnativa (changes requested):
LGTM, just a minor comment. I hope you can consider it.

carlos-lopez-tecnativa (approved):

PR #294, opened 2026-08-21
[18.0][FIX] hr_attendance_report_theoretical_time: Avoid overlap empty attendances

victoralmau:
Avoid overlap empty attendances

If there is an attendance record with a check-out date later than the check-in date, do not create an attendance record with that date.

@Tecnativa TT64274

### DIFF
--- a/hr_attendance_report_theoretical_time/models/hr_employee.py
+++ b/hr_attendance_report_theoretical_time/models/hr_employee.py
@@ -66,8 +66,17 @@ class HrEmployee(models.Model):
                 ]
             )
         )
+        # It's important to take both check_in and check_out into account because
+        # (although strange) there could be an error where check_in=date1 and
+        # check_out=date2; in that case, we don't want to create a record for date2
+        # because they would overlap.
         attendance_dates = sorted(
-            {check_in.date() for check_in in items.mapped("check_in")}
+            {
+                dt.date()
+                for attendance in items
+                for dt in (attendance.check_in, attendance.check_out)
+                if dt
+            }
         )
         dates_to_create = {}
         expected_attendances = self.resource_calendar_id._work_intervals_batch(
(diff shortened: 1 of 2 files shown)


carlos-lopez-tecnativa (commented):
As I understand it, if I have an attendance record created manually from the 3rd to the 6th, the automatic attendances should not be created within this date range, right?

In this example, I see that records are created for the 4th, 5th, and 6th.

Image

Image

victoralmau:
> As I understand it, if I have an attendance record created manually from the 3rd to the 6th, the automatic attendances should not be created within this date range, right?
> 
Exactly—maybe the automatic entry was created BEFORE the attendance was modified or created? The added test checks precisely that use case.

carlos-lopez-tecnativa:
I found two cases. Please let me know if they are valid cases or if I am missing something.

1. Create the employee and save it. Later, set the `Theoretical Hours Start Date`. At this point, the empty attendances are created, but when I create a manual attendance, the overlapping records are not deleted.


https://github.com/user-attachments/assets/7584b6c9-8aa2-4d76-b28e-5b23e11b6db0


2. Create the employee and set the `Theoretical Hours Start Date` before saving. Then, create the attendance. When the cron runs, an error is raised.


https://github.com/user-attachments/assets/630e478d-47b6-4440-873e-0a3164c5fd88

victoralmau:
Looking at your examples, you're “forcing” a use case that I'm not sure actually happens in “day-to-day” operations: creating attendance records for dates prior to today. 

I'll try to explain what you've done and why this PR doesn't (yet) solve that problem.

You create an employee and set their theoretical hours date to x; that will automatically create archived attendance records from that date until today (actually, yesterday). From now on, that employee will create their own attendance records starting today, right?

The use case being addressed in this PR is:
- There are already some attendance records with a start date of 3 and an end date of 5
- The cron job runs > It should not attempt to create archived attendance records for days 4 and 5.

carlos-lopez-tecnativa (approved):
Okay, I understand. Sorry for the confusion.

pedrobaeza (approved):
I understand that the problem is that the system won't let you create that attendances because the constraint of overlapping, but the data isn't correct although this. Later adjustments should be done, but at least we avoid the crash in the cron.

/ocabot merge patch

PR #299, opened 2026-08-26
[18.0][IMP+FIX] hr_attendance_report_theoretical_time: Empty Attendance Records improvements

victoralmau:
Changes done:
- [x] Display the selected archived employee
- [x] Add `widget="daterange"` to Recompute Theoretical Attendances wizard
- [x] Create Empty Attendance Records in the Recompute Theoretical Attendances Wizard
- [x] Avoid errors when editing an employee

Please @carlos-lopez-tecnativa and @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/hr_attendance_report_theoretical_time/models/hr_employee.py
+++ b/hr_attendance_report_theoretical_time/models/hr_employee.py
@@ -50,8 +50,16 @@ class HrEmployee(models.Model):
         directly, so employee attendance records are created for past days
         when they should have clocked in."""
         self.ensure_one()
+        limit_date_from = (
+            self.theoretical_hours_start_date
+            or fields.Datetime.context_timestamp(self, self.create_date).date()
+        )
+        date_from = max(date_from, limit_date_from)
         day_from = datetime.combine(date_from, time.min)
         from_datetime = utc.localize(day_from).astimezone(timezone(self.tz or "UTC"))
+        today = fields.Date.context_today(self)
+        limit_date_to = today - relativedelta(days=1)
+        date_to = min(date_to, limit_date_to)
         day_to = datetime.combine(date_to, time.max)
         to_datetime = utc.localize(day_to).astimezone(timezone(self.tz or "UTC"))
         items = (
--- a/hr_attendance_report_theoretical_time/models/hr_employee.py
+++ b/hr_attendance_report_theoretical_time/models/hr_employee.py
@@ -29,7 +29,7 @@ class HrEmployee(models.Model):
             for item in self:
                 date_from = (
                     item.theoretical_hours_start_date
-                    or fields.Datetime.context_timestamp(item.create_date).date()
+                    or fields.Datetime.context_timestamp(item, item.create_date).date()
                 )
                 attendance_model.sudo().search(
                     [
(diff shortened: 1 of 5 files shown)


hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -27,10 +27,7 @@ def write(self, vals):
         if "theoretical_hours_start_date" in vals:
             attendance_model = self.env["hr.attendance"].with_context(active_test=False)
             for item in self:
-                date_from = (
-                    item.theoretical_hours_start_date
-                    or fields.Datetime.context_timestamp(item.create_date).date()
-                )
+                date_from = item.theoretical_hours_start_date or item.create_date.date()
```
carlos-lopez-tecnativa: I suggested using `context_timestamp` to prevent mistakes in countries where the timezone is behind UTC.

If you review the function signature, it expects the first parameter to be a record and the second parameter to be a datetime. So, please don't remove this conversion; update the parameters instead.

https://github.com/odoo/odoo/blob/cc7b8068441ca804e3e185fffa45c3d01ae694c6/odoo/fields.py#L2434

```suggestion
                date_from = item.theoretical_hours_start_date or fields.Datetime.context_timestamp(item, item.create_date).date()
```

hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -50,8 +47,13 @@ def _action_create_empty_attendance(self, date_from=False, date_to=False):
         directly, so employee attendance records are created for past days
         when they should have clocked in."""
         self.ensure_one()
+        limit_date_from = self.theoretical_hours_start_date or self.create_date.date()
```
carlos-lopez-tecnativa: The same applies here and in other places.

carlos-lopez-tecnativa (changes requested):
LGTM, but I don't want to lose the context_timestamp. Could you please reconsider this?

hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -50,8 +50,16 @@ def _action_create_empty_attendance(self, date_from=False, date_to=False):
...
         when they should have clocked in."""
         self.ensure_one()
+        limit_date_from = (
+            self.theoretical_hours_start_date
+            or fields.Datetime.context_timestamp(self, self.create_date).date()
+        )
+        date_from = max(date_from, limit_date_from)
         day_from = datetime.combine(date_from, time.min)
         from_datetime = utc.localize(day_from).astimezone(timezone(self.tz or "UTC"))
+        today = fields.Date.today()
```
carlos-lopez-tecnativa: Similar case: when you only need the date, take the user's timezone into consideration.
```suggestion
        today = fields.Date.context_today(self)
```

carlos-lopez-tecnativa (changes requested):
I think this module could benefit from a general review of how dates and datetimes are handled with timezones. I see some parts of the code that don't take this into account, but I don't want to take this PR out of scope. Just one more change.

As a general rule, I suggest:

- If you need the current date as a Date value, use `fields.Date.context_today` instead of `fields.Date.today`.
- If you are working with a `Datetime` field and need to extract only the date part while taking the user's timezone into account, use `fields.Datetime.context_timestamp(record, record.datetime_field)` instead of `record.datetime_field.date()`.
- 

Everyone in LATAM will appreciate it.

carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #300, opened 2026-08-27
[18.0][ADD] hr_attendance_employee_calendar_planning: New module

victoralmau:
New module

Related to https://github.com/OCA/hr/pull/1601#issuecomment-5427316023

Locked by:
- [x] `hr_employee_calendar_planning`: https://github.com/OCA/hr/pull/1601

Please @carlos-lopez-tecnativa and @pedrobaeza can you review it?

**Note**: Avoiding manual calls to `_compute_is_flexible()` will be “complicated”; we would have to set the `is_flexible` field to store=False or something similar (I don't like that change).

@Tecnativa TT62028


victoralmau:
Part of this module would not be necessary if, in `hr_employee_calendar_planning`, the `_attendance_intervals_batch()` method returned an empty Intervals object when lunch=True and the calendar is flexible (indirectly, this is the method that is called if the calendar is not flexible: https://github.com/odoo/odoo/blob/d1759c9e6147cc18b7f7f589e669e7afddb58ad6/addons/hr_attendance/models/hr_attendance.py#L240).

carlos-lopez-tecnativa (approved):
LGTM, the case mentioned in this comment has been solved: https://github.com/OCA/hr/pull/1601#issuecomment-5424927774

pedrobaeza (approved):
/ocabot merge nobump

PR #302, opened 2026-08-31
[19.0][IMP] hr_attendance_report_theoretical_time: Add leave hours column to report

victoralmau:
FWP from 18.0: https://github.com/OCA/hr-attendance/pull/290

Add leave hours column to report 

The Leave column is added to the report so that leave data can also be viewed for informational purposes

@Tecnativa TT62028

### DIFF
--- a/hr_attendance_report_theoretical_time/README.rst
+++ b/hr_attendance_report_theoretical_time/README.rst
@@ -45,12 +45,12 @@ types if specified in them.
 As an example, imagine a work week with 40 theoretical hours, and these
 attendance situation:
 
-- Monday: Worked 10 hours
-- Tuesday: Worked 10 hours
-- Wednesday: Worked 10 hours
-- Thursday: Worked 10 hours
-- Friday: Ask for a compensation leave (said leave type), as already
-  worked 40 hours.
+-  Monday: Worked 10 hours
+-  Tuesday: Worked 10 hours
+-  Wednesday: Worked 10 hours
+-  Thursday: Worked 10 hours
+-  Friday: Ask for a compensation leave (said leave type), as already
+   worked 40 hours.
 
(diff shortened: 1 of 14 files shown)


carlos-lopez-tecnativa (changes requested):
LGTM, but I noticed that this field does not have a widget to format the float as a time, which could cause the values to be interpreted incorrectly. Please add the corresponding `widget="float_time"`

Please also check V18, as the widget is necessary there too.

Image

carlos-lopez-tecnativa (approved):

pedrobaeza (changes requested):
I'm now thinking that both `leave_hours` and `theoretical_hours` require a pre-init-hook for pre-creating them, and this migration script should cover the case that you come from 18 and the column already exists.

victoralmau:
> I'm now thinking that both `leave_hours` and `theoretical_hours` require a pre-init-hook for pre-creating them, and this migration script should cover the case that you come from 18 and the column already exists.

So I create a `pre_init_hook` to create the columns if they don't exist (this wasn't currently done in any version for the existing `theoretical_hours` column). Ok, **but**... what initialization value do I set using SQL?  In my opinion, we should let the compute methods run and set the appropriate value for each record, so I don't understand the need for the `pre_init_hook`.

pedrobaeza:
The reason is simple: if you have a lot or attendances records, you will freeze your Odoo instance installing the module.

victoralmau:
I understand that (something that already happens in previous versions), but the question is what initial value is set. 
Perhaps (this is just a guess) you want to set a value of 0 in both columns and then have the data recalculated when needed.

pedrobaeza:
Yes, that's the best guess, as this module AFAIK has a tool to recompute past attendances.

pedrobaeza (approved):
/ocabot merge nobump

PR #303, opened 2026-09-01
[18.0][FIX] hr_attendance_employee_calendar_planning: Fix check_in/check_out dates

victoralmau:
Fix check_in/check_out dates

```
File "/opt/odoo/auto/addons/hr_employee_calendar_planning/models/hr_employee.py", line 57, in <lambda>
    lambda x: (not x.date_start or (from_date and x.date_start <= from_date))
TypeError: can't compare datetime.datetime to datetime.date
```
Use case example:
- Set a flexible calendar for an employee (e.g., Admin)
- Click the button in the upper-right corner to check in

@Tecnativa TT64335

### DIFF
--- a/hr_attendance_employee_calendar_planning/models/hr_attendance.py
+++ b/hr_attendance_employee_calendar_planning/models/hr_attendance.py
@@ -30,8 +30,8 @@ class HrAattendance(models.Model):
         for employee, attendance_dates in employee_attendance_dates.items():
             data_item = defaultdict(set)
             data_item[employee] = attendance_dates
-            start_date = min(check_in for check_in, _check_out in attendance_dates)
-            end_date = max(check_out for _check_in, check_out in attendance_dates)
+            start_date = min(day for _datetime_date, day in attendance_dates)
+            end_date = max(day for _datetime_date, day in attendance_dates)
             items = self.filtered(
                 lambda x, employee=employee: x.exists() and x.employee_id == employee
             )
--- a/hr_attendance_employee_calendar_planning/models/hr_attendance.py
+++ b/hr_attendance_employee_calendar_planning/models/hr_attendance.py
@@ -1,7 +1,7 @@
 # Copyright 2026 Tecnativa - Víctor Martínez
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


hr_attendance_employee_calendar_planning/models/hr_attendance.py
```diff
@@ -30,8 +31,14 @@ def _update_overtime(self, employee_attendance_dates=None):
         for employee, attendance_dates in employee_attendance_dates.items():
             data_item = defaultdict(set)
             data_item[employee] = attendance_dates
-            start_date = min(check_in for check_in, _check_out in attendance_dates)
-            end_date = max(check_out for _check_in, check_out in attendance_dates)
+            start_date = min(
+                check_in.date() if isinstance(check_in, datetime) else check_in
```
pedrobaeza: There can be problems with timezones in this case. Why there's a mix of date/datetimes?

victoralmau: It has already been changed (using the https://github.com/odoo/odoo/blob/4cadc97d771824843449550b31250220b856218d/addons/hr_attendance/models/hr_attendance.py#L306 method) to use the existing date.

hr_attendance_employee_calendar_planning/models/hr_attendance.py
pedrobaeza: I'm seeing that here there's the same problem converting datetime to dates without being aware of timezones.

victoralmau: We will address the details of this method in another PR to try to clarify it further, as it will require a change to `hr_employee_calendar_planning`.

victoralmau: Changed: The solution at https://github.com/OCA/hr-attendance/pull/304 doesn't work.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

victoralmau:
Can you try the merge process again?

PR #307, opened 2026-09-02
[19.0][MIG] hr_attendance_employee_calendar_planning: Migration to 19.0

victoralmau:
Migration to 19.0

Locked by: 
- [x] `hr_employee_calendar_planning`: https://github.com/OCA/hr/pull/1618
- [x] `hr_employee_calendar_planning`: https://github.com/OCA/hr/pull/1619
- [x] `hr_employee_calendar_planning`: https://github.com/OCA/hr/pull/1620
- [x] `hr_attendance_report_theoretical_time`: https://github.com/OCA/hr-attendance/pull/311

@Tecnativa


carlos-lopez-tecnativa (commented):
Is this module still necessary in V19?

I tested without this module and repeated the same case I mentioned here: https://github.com/OCA/hr/pull/1601#issuecomment-5424927774, and it worked.

victoralmau:
This module is required for the method to be overridden to correctly handle this condition: https://github.com/odoo/odoo/blob/65dae4abd4a743e1e1b9a7e9b0751af0815908a7/addons/hr_attendance/models/hr_attendance.py#L192.

I'm going to improve the tests to better confirm that without this method, a specific use case would not work.

victoralmau:
It is not necessary to override the `_get_worked_hours_in_range()` method in `hr.attendance` because even if the code reaches https://github.com/odoo/odoo/blob/65dae4abd4a743e1e1b9a7e9b0751af0815908a7/addons/hr_attendance/models/hr_attendance.py#L193 and then at https://github.com/OCA/hr/blob/37cc00ed8aa56ce492c078463ac221a86afdfc19/hr_employee_calendar_planning/models/resource_calendar.py# L346, the appropriate context is defined to access https://github.com/odoo/odoo/blob/f586c1dad0b9a7653c45d1f82a9c3989a287c641/addons/resource/models/resource_calendar.py#L337

carlos-lopez-tecnativa (changes requested):
What is your use case for this module now?

I tested the same example that didn't work in V18 and required this module, but the same example now works without it in V19. So, could you please provide an example of a use case that still requires this module, and confirm whether you can reproduce it without the module?

Additionally, I found an issue that is not directly related to this module, but is indirectly related through its dependencies. Please review this.

In the overtime rule lines, there is a Many2one field related to `resource.calendar` that has a domain.

Select `If the employee works?` as `Outside of a specific schedule`. Then, try to select a calendar in the `Schedule` field. A traceback is raised because the field has a domain that uses the `flexible_hours` field. This field is stored in core, but you set it to `store=False` in [hr_employee_calendar_planning](https://github.com/OCA/hr/blob/37cc00ed8aa56ce492c078463ac221a86afdfc19/hr_employee_calendar_planning/models/resource_calendar.py#L59).

Image

Image

victoralmau:
> What is your use case for this module now?
> 

The use case is the test for this module.

ejemplo-1
ejemplo-2

When creating an attendance (with a start and end date), an incorrect `overtime_hours` value is set (if the calendar is 4 hours per day and the employee only worked 2 hours, the value should be 0).

ejemplo-3

These are the cases covered in the test.

> I tested the same example that didn't work in V18 and required this module, but the same example now works without it in V19. So, could you please provide an example of a use case that still requires this module, and confirm whether you can reproduce it without the module?
> 
> Additionally, I found an issue that is not directly related to this module, but is indirectly related through its dependencies. Please review this.
> 
> In the overtime rule lines, there is a Many2one field related to `resource.calendar` that has a domain.
> 
> Select `If the employee works?` as `Outside of a specific schedule`. Then, try to select a calendar in the `Schedule` field. A traceback is raised because the field has a domain that uses the `flexible_hours` field. This field is stored in core, but you set it to `store=False` in [hr_employee_calendar_planning](https://github.com/OCA/hr/blob/37cc00ed8aa56ce492c078463ac221a86afdfc19/hr_employee_calendar_planning/models/resource_calendar.py#L59).
> Image Image

This bug will be fixed in v18 with https://github.com/OCA/hr/pull/1616 (once it's merged, I'll add it to v19).

carlos-lopez-tecnativa (approved):
It's curious, but when I tested it in the HR repository with both hr_employee_calendar_planning and hr_attendance installed, it didn't work. The hr_attendance_employee_calendar_planning module doesn't exist in the HR repository:
http://oca-hr-19-0-23e948031779.runboat.odoo-community.org/odoo/action-223/18

However, when I tested it in the Runboat instance of this PR, without hr_attendance_employee_calendar_planning installed, it worked. But why? If this module isn't installed, changes in this module shouldn't affect the logic:
http://oca-hr-attendance-19-0-pr307-aa3b80ba1a51.runboat.odoo-community.org/odoo/attendances/20
Anyway, LGTM.

pedrobaeza (approved):
/ocabot migration hr_attendance_employee_calendar_planning
/ocabot merge nobump

PR #310, opened 2026-09-04
[18.0][IMP] hr_attendance_report_theoretical_time: hr_holidays_public compatibility

victoralmau:
`hr_holidays_public` compatibility

@Tecnativa

### DIFF
--- a/hr_attendance_report_theoretical_time/models/hr_employee.py
+++ b/hr_attendance_report_theoretical_time/models/hr_employee.py
@@ -103,7 +103,13 @@ class HrEmployee(models.Model):
             }
         )
         dates_to_create = {}
-        expected_attendances = self.resource_calendar_id._work_intervals_batch(
+        calendar = self.resource_calendar_id.with_context(
+            # It is important to define these context values so that
+            # `hr_holidays_public` excludes the corresponding public holidays
+            exclude_public_holidays=True,
+            employee_id=self.id,
+        )
+        expected_attendances = calendar._work_intervals_batch(
             from_datetime,
             to_datetime,
             resources=self.resource_id,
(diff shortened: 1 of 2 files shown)


hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -103,7 +103,13 @@ def _action_create_empty_attendance(self, date_from=False, date_to=False):
             }
         )
         dates_to_create = {}
-        expected_attendances = self.resource_calendar_id._work_intervals_batch(
+        calendar = self.resource_calendar_id.with_context(
+            # It is important to define these context values so that
+            # `hr_holidays_public` excludes the corresponding public holidays
+            exclude_public_holidays=True,
```
pedrobaeza, line 109: But this depends on the type of leave, so for the column `theoretical_hours` this not always like that. Or am I wrong?

victoralmau: I think it's important to clarify what's happening: inactive attendance records are created because the system considers that the employee is supposed to work on that day. Is that correct? In my opinion, no—that's why this correction is being added.

Having clarified that, I'm not exactly sure what that has to do with the `theoretical_hours` column (I didn't mention it).

pedrobaeza: Well, I would appreciate this kind of explanations in the commit message from the beginning to understand the kind of issue you are trying to fix. So on a public holiday no attendance record should exist? Shouldn't it contain any kind of data in `theoretical_hours` nor `leave_hours` fields?

victoralmau: That commit message again... I added a comment in the code at the specific location to indicate that public holidays should be excluded, but that isn't enough; now I'll add a very detailed explanation to the commit message to clarify all of this.

Let me clarify again (in my opinion): if a given day is a public holiday, an inactive attendance record should not be generated because the employee does not work that day. Absences should also be 0 (although they currently aren’t because public holidays aren’t being excluded with the appropriate context).

pedrobaeza: As you know, if I ask, it's because I don't see it clear. Maybe it's my fault, but think that others may have similar doubts in any moment. That's why I always ask for being the clearest possible on the explanations, but OK, I don't insist more.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

PR #312, opened 2026-09-08
[19.0][IMP] hr_attendance_report_theoretical_time: hr_holidays_public compatibility

victoralmau:
FWP from 18.0: https://github.com/OCA/hr-attendance/pull/310

`hr_holidays_public` compatibility

The hr_attendance_report_theoretical_time module depends on hr_holidays_public; the correct context is currently taken into account at https://github.com/OCA/hr-attendance/blob/a0cc683e4b17639c8dd4c32c5cb2f4bfb9cbad0e/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py#L117 so that the value of the `theoretical_hours` column is correct (0); since the automatic creation of archived attendance records on employees’ workdays was added at https://github.com/OCA/hr-attendance/pull/247, it is not at all necessary to create archived attendance records if the day in question is a public holiday; therefore, the appropriate context is added in the specific location to prevent the record from being created unnecessarily.

https://github.com/Tecnativa

### DIFF
--- a/hr_attendance_report_theoretical_time/models/hr_employee.py
+++ b/hr_attendance_report_theoretical_time/models/hr_employee.py
@@ -103,7 +103,13 @@ class HrEmployee(models.Model):
             }
         )
         dates_to_create = {}
-        expected_attendances = self.resource_calendar_id._work_intervals_batch(
+        calendar = self.resource_calendar_id.with_context(
+            # It is important to define these context values so that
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Thanks for the extra explanations.

carlos-lopez-tecnativa (approved):

PR #317, opened 2026-09-17
[17.0][IMP] hr_attendance_reason: make the reason name translatable

rrebollo:
`hr.attendance.reason.name` is user facing: it is rendered in the kiosk reason selector, in the systray check-in/out popup and in the attendance form. Without `translate=True`, every non-English user sees the reason names in English — we hit this in a Spanish deployment, where the kiosk offered the reason untranslated.

The analogous core model `hr.leave.type` already declares its `name` as translatable; `hr.attendance.reason` is the outlier.

No migration script is needed: `Char._convert_db_column` detects the change and delegates to `convert_column_translatable`, which wraps the existing value as the `en_US` translation, so existing reasons keep their name after the upgrade. Verified on a live upgrade: the column becomes `jsonb` and the seeded reasons keep their values as `{"en_US": "..."}`.

@BinhexTeam T12289


pedrobaeza (approved):

szalatyzuzanna (approved):
LGTM!

pedrobaeza:
Please fw-port it to upper versions:

/ocabot merge patch

